PR #3, opened 2017-06-09
Routing and templates

rviscomi:
Making this more app-like.

The list of possible reports are maintained in a JSON config file. The /reports page dynamically generates the report links based on the metrics in the config file.

Later, I'll use the metric from the route to load the appropriate data file or show an error. For now there's only one data file, so all reports show the same thing.

"metric" probably isn't the best name, but I'll change that as needed.


templates/report.html
```diff
@@ -1,305 +1,74 @@
...
+{% block content %}
+	<img id="logo" src="/static/img/ha.png"/>
+
+	<div id="container" style="min-width: 310px; height: 400px; margin: 0 auto"></div>
+
+	<table id="desktop"></table>
+	<table id="mobile"></table>
+
+	<script src="https://code.highcharts.com/highcharts.js"></script>
+	<script src="http://code.highcharts.com/modules/exporting.js"></script>
```
igrigorik: https? :-)

PR #7, opened 2017-08-07
npm, webpack, babel

rviscomi:
Reports page is still WIP but I want to get this branch merged with master because it includes significant development improvements including npm, webpack, and babel.


igrigorik:
Haven't done a thorough scrub the JS+python, but on a quick pass this looks good.

One question:  static/json/* --> are we actually planning to version and store all the data in Git? Should we exclude these files? Could we map those files into a separate (cnamed?) bucket that lives on Google Storage?

rviscomi:
Yes, I'd like to store the JSON data on Google Storage. Probably also move the reports.json manifest there too and keep it in sync when new data is added.

PR #17, opened 2018-02-13
prefer BQ datasubsets over har/runs datasets

rviscomi:


rviscomi:
See also https://github.com/HTTPArchive/bigquery/pull/23

Issue #13, opened 2018-01-03
🎓 Graduate to the default version of HTTP Archive


rviscomi:
# Roadmap

The goal is for beta.httparchive.org to become the new httparchive.org, targeting some time in Q1 2018. The following tasks are launch blocking:

- [x] automate the report generation after new tables are added to BigQuery
- [x] resolve all placeholder graphics (eg report images)
- [x] write descriptions for all reports/metrics
- [x] rewrite FAQ section
- [x] fix UX issues with data visualization
- [x] implement a redirect solution for legacy URLs
- [x] redirect to HTTPS
- [x] reorganize GitHub repositories

## Report automation

The reports are [manually generated](https://github.com/HTTPArchive/beta.httparchive.org/tree/master/sql) with scripts. These scripts should be automated to run whenever a crawl is complete. Some subtasks for this issue:

- [x] Queries may depend on convenience datasets (eg `httparchive.lighthouse`) which are manually copied from catch-all datasets (eg `httparchive.har`). New tables in these datasets must all be created automatically.
- [x] Use pubsub or similar tools to trigger report jobs after the dataflow job has completed successfully.

## Graphics

Concept graphics for the JS report:

![image](https://user-images.githubusercontent.com/1120896/34532420-13aa494a-f084-11e7-8d37-e18192dfd59a.png)

We would need similar graphics for each report. It would also be nice to have a default graphic for new reports until a permanent one could be made.

## Written descriptions

For example, "Total KB" should have a description like `The sum of transfer size kilobytes of all resources requested by the page.`.

Reports should describe their contents and maybe even a brief analysis of the overall trends.

## Rewrite FAQs

The legacy FAQs are somewhat outdated. The new FAQ page should contain updated information including any new content related to the new reports/metrics/visualizations.

## Data viz UX

Some feedback on the charts include:

- what the heck is a CDF/PDF?
- make the tooltips more descriptive
- unclear what the outlier bin is in the histograms
- not obvious how to switch between timeseries/histogram modes or that separate modes exist
- collapse desktop and mobile tables into one with both histograms side by side

## Legacy redirects

[...]

PR #18, opened 2018-03-05
FAQ

rviscomi:
Demo: https://beta.httparchive.org/faq

Features:
- markdown-supported config file of questions and answers
  - server-side rendered FAQ template
  - automatic permalink generation of questions
- takes many of the relevant FAQs from the [legacy page](http://httparchive.org/about.php#faq)
- updated answers
  - eg BigQuery answer contains links to modern [docs](https://github.com/HTTPArchive/httparchive/blob/master/docs/bigquery-gettingstarted.md) and resources
- changelog is dynamically generated based on [JSON file](https://raw.githubusercontent.com/HTTPArchive/httparchive/master/docs/changelog.json) maintained in GitHub
  - supports URL anchors to jump to change (non-trivial when it's dynamically added to the DOM after load)
  - see comments in the code about HTML support in changelog (there isn't any)

Progress on #13


src/js/faq.js
```diff
@@ -0,0 +1,52 @@
...
+		}).forEach(change => {
+			const date = el('dt');
+			date.id = `change-${change.date}`;
+			anchors.add(`#${date.id}`);
+			const fullDate = getFullDate(new Date(change.date));
+			date.innerText = `${fullDate}: ${change.title}`;
+			dl.appendChild(date);
+			const desc = el('dd');
+			// This description may contain URLs.
+			// We are intentionally NOT converting them to clickable anchors
```
rviscomi: One secure alternative may be to allow reference URLs in the changelog object. For example:

```
{
  date: 123456,
  title: 'Foo',
  desc: 'Foo bar.',
  url: 'example.com'
}
```

Then here in the generator we can create an anchor tag if `change.url` exists, with the href and innerText values set to that value.

faq.py
```diff
@@ -0,0 +1,114 @@
+from collections import OrderedDict
+
+CHANGELOG_QUESTION = 'What changes have been made to the test environment that might affect the data?'
+
+def get_anchor(question):
+    return question.lower().replace(' ', '-').replace('?', '').replace('\'', '')
+
+# Keys = Questions
+# Values = Markdown-supported answers
```
igrigorik: If we're writing markdown, could we pull these out into a separate .md file for easier editing? Seems a bit odd and awkward to maintain markdown in Python strings?

faq.py
```diff
@@ -0,0 +1,114 @@
...
+Make sure to send a follow-up email to 
+[donations@archive.org](mailto:donations@archive.org?subject=HTTP+Archive+donation) 
+designating your donation to the "HTTP Archive".
+'''
+
+FAQ['Who maintains the HTTP Archive?'] = \
+'''[Steve Souders](https://twitter.com/Souders) created the HTTP Archive in 2010.
+It's built on the shoulders of [Pat Meenan](https://twitter.com/patmeenan)'s 
+[WebPageTest](https://webpagetest.org/) system. 
+We've received patches from several individuals including Jonathan Klein, Yusuke Tsutsumi, Carson McDonald, James Byers, Ido Green, Charlie Clark, Jared Hirsch, and Mike Pfirrmann.
```
igrigorik: To make sure this list is up to date, could we link to the contributions page on GitHub? Makes everyone's life simpler?

src/js/faq.js
igrigorik: 👍 

However, if we're already generating the FAQ page on the server, why bother with the client-side stuff here and not also parse and generate the changelog contents in python?

rviscomi: Generating the changelog on the server doesn't allay the XSS concerns. We'd still need to decide whether to render the content as text or markup, the only difference being in the jinja template rather than the JS.

Also, the changelog is an external dependency on GitHub, so fetching it in the client at least allows the page to start rendering while it loads. It also avoids the reimplementing the backend load/cache/invalidate logic of an HTTP resource.

faq.py
rviscomi: 👍

src/js/faq.js
igrigorik: Fair enough. I like the explicit url pattern. It does limit us to just one URL in the details, but that's probably more than enough — e.g. a pointer to a bug, or discussion elsewhere.

faq.py
rviscomi: I agree it'd be way easier to write markdown in a md file as opposed to raw strings.

Do you have any suggestions to make that work with the autogenerated question permalinks? Best I can think of ATM is lots of string splits to get at each Q & A in a format like this file, making lots of assumptions about use of whitespace and hierarchical (h3) markdown.

src/js/faq.js
rviscomi: Agreed. We can add fanciness if needed, like key/value pairs of label/URL to handle multiple named links. It can appear below the changelog entry as a "See also: X, Y" list of links, kind of similar to how we [link reports with overlapping metrics](https://beta.httparchive.org/reports/state-of-the-web#bytesTotal).

faq.py
igrigorik: We can create a convention around each section being an h2/h3 with paragraph answer after it. 

For the slugs, did you look at existing solutions? A quick search yields..
- https://github.com/Flet/github-slugger
- https://github.com/jonschlinkert/markdown-toc

rviscomi: Yeah declaring a convention (and enforcing in code reviews) would work. I'll work on a solution and ping this thread when it's ready to look at. Or LMK if it's non-blocking and I can write a TODO and file an issue.

src/js/faq.js
rviscomi: Done

faq.py
igrigorik: I wouldn't hold the launch on this, but personally I would like to see the text content living in a separate file. I'll let you make the call.

rviscomi: Updated to have a separate markdown file. This is also nice because it is renderable on GitHub. See https://github.com/rviscomi/beta.httparchive.org/blob/8a8d39ae17cfcb10793d7351267d3b90e4a070f2/docs/faq.md

No other third parties were needed. I used a regex to find all headings, and map the questions to slugs. I used a similar client-side technique as the changelog to dynamically add anchors to the questions and scroll when needed. The FAQ itself is rendered server-side without the anchors built-in.

rviscomi:
PTAL (see https://github.com/HTTPArchive/beta.httparchive.org/pull/18#discussion_r172662088 which is collapsed by default 😞)

PR #19, opened 2018-03-06
Metric and report descriptions

rviscomi:
Features:
- Descriptions added to all metrics
- Markdown supported
- "State of the Web" selected as featured report (why not)

cc @ekharvey. @pmeenan, I'd especially like your help describing the `_cpu` metrics from WPT.

Progress on https://github.com/HTTPArchive/beta.httparchive.org/issues/13


config/reports.json
```diff
@@ -8,51 +8,59 @@
...
 			"name": "Font Bytes",
-			"type": "KB"
+			"type": "KB",
+			"description": "The sum of transfer size kilobytes of all fonts requested by the page. A font is identified as a resource with the `eot`, `ttf`, `woff`, `woff2`, or `otf` file extensions or a MIME type containing `font`."
 		},
 		"bytesHtml": {
 			"name": "HTML Bytes",
-			"type": "KB"
+			"type": "KB",
+			"description": "The sum of transfer size kilobytes of all HTML documents requested by the page. An HTML document is identified as a resource with the `html` or `html` file extensions or a MIME type containing `html`."
```
igrigorik: Curious, where do we slot text/plain? :)

config/reports.json
```diff
@@ -8,51 +8,59 @@
...
 			"name": "Other Bytes",
-			"type": "KB"
+			"type": "KB",
+			"description": "The sum of transfer size kilobytes of all unidentified resources requested by the page. An unidentified resource is one that does not match a known script, font, image, stylesheet, XML file, video, HTML document, or plaintext file."
 		},
 		"bytesTotal": {
 			"name": "Total Kilobytes",
 			"type": "KB",
-			"description": "The sum of <a href=\"https://www.w3.org/TR/resource-timing-2/#dom-performanceresourcetiming-transfersize\">transfer size</a> kilobytes of all resources requested by the page.",
+			"description": "The sum of [transfer size](https://www.w3.org/TR/resource-timing-2/#dom-performanceresourcetiming-transfersize) kilobytes of all resources requested by the page.",
```
igrigorik: We only link it here, I assume this is intentional and bytesTotal will be shown as one of the first graphs?

config/reports.json
```diff
@@ -8,51 +8,59 @@
...
 			"timeseries": {
 				"redundant": true
 			}
 		},
 		"bytesVideo": {
 			"name": "Video Bytes",
 			"type": "KB",
+			"description": "The sum of transfer size kilobytes of all videos requested by the page. A video is identified as a resource with the `mp4`, `swf`, `f4v`, or `flv` file extensions or a MIME type containing `flash`.",
```
igrigorik: What about WebM, etc?

Meta question: where are these definitions coming from? Could we simplify all this by pointing to the source?

config/reports.json
```diff
@@ -80,18 +90,20 @@
 		"fmp": {
 			"name": "First Meaningful Paint",
```
igrigorik: As an aside, we may want to remove this.. We're moving away from FMP (for now, at least) in LH. Instead show FCP to align with CrUX and RUM apis.

config/reports.json
```diff
@@ -219,6 +243,7 @@
 		"cruxDcl": {
 			"name": "DOMContentLoaded",
 			"type": "seconds",
+			"description": "The number of seconds from the time the navigation started until the browser first renders any text or graphics. This metric comes from real user data in the [Chrome UX Report](https://developers.google.com/web/tools/chrome-user-experience-report/). See the [Paint Timing API](https://w3c.github.io/paint-timing/#first-contentful-paint) for more info.",
```
igrigorik: Drop the paint timing bits.. :)

config/reports.json
```diff
@@ -229,6 +254,7 @@
 		"cruxOl": {
 			"name": "Onload",
 			"type": "seconds",
+			"description": "The number of seconds from the time the navigation started until the document and all of its dependent resources have finished loading. This metric comes from real user data in the [Chrome UX Report](https://developers.google.com/web/tools/chrome-user-experience-report/). See the [Paint Timing API](https://w3c.github.io/paint-timing/#first-paint) for more info.",
```
igrigorik: Ditto, drop the paint timing bits.. :)

config/reports.json
rviscomi: > WebM

While going through the source code for these metrics, I definitely got the feeling that some were not up to date. I'm happy to take a TODO to modernize the metrics, while migrating from the old PHP/MySQL code to BigQuery.

All bytes/requests metrics are defined here: https://github.com/HTTPArchive/httparchive/blob/b52c778f67f7d3217e5c1e763372d4ee0a66e5a3/bulktest/batch_lib.inc#L933-L973

config/reports.json
rviscomi: Yeah the descriptions were already feeling redundant and I didn't want to repeat the same link 8 times 😛 

I rearranged the metrics so that the totals are first.

config/reports.json
rviscomi: That would be categorized as "text". See https://github.com/HTTPArchive/httparchive/blob/b52c778f67f7d3217e5c1e763372d4ee0a66e5a3/bulktest/batch_lib.inc#L967-L970

Not currently tracking it in these reports though.

config/reports.json
igrigorik: Yikes, yeah.. that needs some work. Doesn't WPT already have this sort of logic somewhere? I'd be great if we could avoid having to reinvent the wheel. /cc @pmeenan

config/reports.json
rviscomi: Done.

Writing down some notes about the switchover here before I forget:

- FCP is not available in Lighthouse, but WPT does expose it in the HAR as `_chromeUserTiming.firstContentfulPaint`
- renamed and rewrote fmp.sql histogram/timeseries queries to use above FCP metric
- renamed the metric in the report.json config and updated its description
- went through the timeseries data to find the first complete crawl with FCP data (2016_12_15) and set the minDate in the metric config
- script to backfill FCP timeseries data: `sql/generateReport.sh -d "fcp.json"`
- script to backfill FCP histogram data: `sql/getBigQueryDates.sh pages '' 2016_12_15 | xargs -I date sql/generateReport.sh -d date/fcp.json`

config/reports.json
rviscomi: Either way, I'm going to be rewriting all this categorization logic as we move off MySQL so I don't think we should block on it now. When we get there, I do agree that it's worth updating the definitions.

LMK if there's anything else in this PR that needs attention.

igrigorik: sg. should we file a bug for this, so we can keep track?

rviscomi: Filed https://github.com/HTTPArchive/beta.httparchive.org/issues/23

rviscomi:
Any other comments, or is this ok to merge?

Issue #25, opened 2018-03-12
Avoid "Infinity%" when timeseries starts at 0


rviscomi:
![image](https://user-images.githubusercontent.com/1120896/37315037-d8f8edda-262d-11e8-8347-ab166934292b.png)

Select starting point as first non-zero value to avoid zero division chaos.

![image](https://user-images.githubusercontent.com/1120896/37315073-109129ba-262e-11e8-875b-5d4e84b2dcf0.png)

PR #26, opened 2018-03-15
Data Viz

rviscomi:
Features:
- [x] simplified timeseries
  - improved tooltip
- [x] fixes [infinity bug (#25)](https://github.com/HTTPArchive/beta.httparchive.org/issues/25)
- [x] simplified histogram
  - outliers > 95% CDF are hidden by default
  - CDF hidden from legend and tooltip
- [x] show/hide CDF when desktop/mobile histogram series shown/hidden
- [x] the singularity is here: eg fix "1 libraries" grammar to "1 library"
- [x] [security policy](https://github.com/GoogleCloudPlatform/flask-talisman) (forced https, csp, hsts, etc)

Progress on #13

image

image

image


rviscomi:
@igrigorik ping -- mostly just looking for feedback on the data viz tooltips, no need for a gritty code review

You can test drive these changes on the staging app: https://httparchive-staging.appspot.com/

igrigorik:
Nice work Rick! One thing that is lost in this update is the CDF values on the histogram..

image

Perhaps this is counter to what you're aiming for here (simplify), but it would be nice if we could somehow persist / communicate that info in the histogram tooltip. WDYT?

rviscomi:
Yeah I had a hard time thinking of a way to include it cleanly without adding confusion.

Maybe use the summation symbol ∑ to indicate CDF, in the same kind of design that we use to show change over time in the timeseries summaries:

image

image


WDYT?

igrigorik:
∑ is not commonly understand symbol for CDF, and afaik I don't think there is one. The best I can come up with is still stick with "CDF: x" but in small text.. =/

rviscomi:
Here's a revision that includes CDF. Using the word "Cumulative" here to avoid jargon. WDYT?
image

igrigorik:
^^ 👍 👍 

Ship it.

rviscomi:
Woohoo! 5/8 of the [graduation checklist](https://github.com/HTTPArchive/beta.httparchive.org/issues/13) completed! 🎉

PR #31, opened 2018-03-22
dynamic dates from GCS

rviscomi:
Features:
- dates.json is no longer manually maintained
- dates are determined at runtime by inspecting available GCS subdirs
- remove obsolete docs
- standardize on tabs everywhere

Progress on https://github.com/HTTPArchive/beta.httparchive.org/issues/13


rviscomi:
@igrigorik this is the last piece of the report automation. GCE will generate the reports after each crawl and GAE will ping GCS periodically for the latest dates.

Sorry to mix this with spaces->tabs, harder to read the diff.

dates.py
```diff
@@ -0,0 +1,18 @@
+import re
+from google.cloud import storage
+
+GCS_BUCKET = 'httparchive'
+
+def get_dates():
```
igrigorik: 👍

PR #35, opened 2018-03-24
WebPageTest integration: Plumbing and simple metric support

rviscomi:
Features:
- documentation updates
  - `wpt` config support
  - README support for authenticating GCP user
- accepts `wptid` URL parameter whose value is the WebPageTest ID
  - need to add appropriate UI to add/remove WPT ID (will initially be an easter egg)
- async fetches WPT results
- extracts values for supported metrics:
  - bytesTotal
  - dcl
  - fcp
  - ol
  - tcp
  - ttfi
  - reqTotal
  - cruxFp
  - cruxFcp
  - cruxDcl
  - cruxOl
- displays values alongside metric summaries

image

In the screenshot above, we can see that the test page has an onload time of 9.8 seconds, which is in the ~90th percentile for desktop and ~80th percentile for mobile.

Progress on #36


config/reports.json
```diff
@@ -213,6 +232,10 @@
 			"description": "The number of seconds from the time the navigation started until the layout is stabilized, web fonts are visible, and the page is responsive to user input. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/consistently-interactive) as is only available in mobile tests.",
 			"histogram": {
 				"minDate": "2017_06_01"
+			},
+			"wpt": {
+				"path": "median$firstView$FirstInteractive",
```
rviscomi: @pmeenan FYI here's an example of how I extract data from the WPT API results. In this case I'm extracting TTFI. But the metric above (TTCI) doesn't seem to have a corresponding WPT value. LMK if any of these paths look off.

pmeenan: The TTI metrics (FirstInteractive  and TimeToInteractive) are not always there because it need 5 seconds of main-thread idle and not all tests go idle for that long at the end.  There is a LastInteractive metric that WPT uses as a fallback when the others are not available and it is the earliest that TTI or first interactive could be if enough idle time would have happened in the test.

In the UI it shows up as First Interactive > <time>

rviscomi: Gotcha thanks! If the metric isn't available we degrade gracefully and omit it from the UI, so that shouldn't be an issue.

rviscomi:
@igrigorik yes, this is on the checklist for https://github.com/HTTPArchive/beta.httparchive.org/issues/36

rviscomi:
looks good?

Issue #33, opened 2018-03-23
Subdomains: Make beta the new www


rviscomi:
Progress on legacy redirects in #13.

- [x] Add an A record to the DNS for the `legacy` subdomain to resolve to the legacy web server
- [x] Add support for the `legacy` subdomain in the legacy web server's nginx and apache configs
- [x] Add SSL cert for the `legacy` subdomain
  - assigned to @pmeenan in https://github.com/HTTPArchive/httparchive/issues/132
- [x] Modify the beta server's 404 page to suggest the same URL on the legacy subdomain
- [x] Replace the `@` and `www` DNS A records to CNAME records, resolving to the new (beta) App Engine server at `ghs.googlehosted.com`
  - do this 4+ hours prior to launch to ensure that it propagates to DNS caches
- [x] Update all internal links to avoid 404s
  - eg stats/trends links on the forum

DNS settings:
- A		216.239.32.21
- A		216.239.34.21
- A		216.239.36.21
- A		216.239.38.21
- AAAA		2001:4860:4802:32::15
- AAAA		2001:4860:4802:34::15
- AAAA		2001:4860:4802:36::15
- AAAA		2001:4860:4802:38::15
- CNAME	ghs.googlehosted.com (www)

rviscomi:
The legacy subdomain is up and running at http://legacy.httparchive.org/. SSL cert pending.

Sanity check: after replacing the DNS records to point to GAE it shouldn't matter that the legacy web server is still configured to handle @ and www subdomains, right? Because the DNS records (after 1 TTL) will never point to that server anyway. So is it necessary to clean up those configs or can they be left indefinitely?

igrigorik:
No harm in keeping existing configuration — it's a noop.

pmeenan:
legacy cert is complete - https://legacy.httparchive.org/

rviscomi:
Added a TODO item to update internal links. Noticed on the forum that we're linking directly to stats/trends on the legacy site: 
![image](https://user-images.githubusercontent.com/1120896/37869684-7fb5aa38-2f92-11e8-944a-250367bdf19d.png)

rviscomi:
Reduced the DNS TTL to 1h last night so no need to wait the full 4 hours now.

A, AAAA, and CNAME records for the root and www subdomain have been set.

rviscomi:
Updated forum template to point to new (HTTPS) URLs.

[...]

PR #37, opened 2018-03-24
link 404 page to legacy site

rviscomi:
Features:
- higher res logo, with gradient "p"
- custom 404 error page with link to legacy subdomain
  - the link URL is the current page's URL with the host replaced by `legacy.httparchive.org`
- adjusted responsiveness of error page template

eg when visiting /trends.php on the new site, the user would see:

image

Progress on #33 and #13


rviscomi:
Done!

Only big thing I changed is mapping `/interesting.php` to the State of the Web report's histogram view ([/reports/state-of-the-web?start=latest](https://beta.httparchive.org/reports/state-of-the-web?start=latest)). Looking at Google Analytics, these redirects should handle 2/3 of page views. We can also consider adding 302 redirects for unsupported pages like [/websites.php](https://legacy.httparchive.org/websites.php) or [/downloads.php](https://legacy.httparchive.org/downloads.php) to their legacy URL equivalents.

Also did the cookie thing with the banner:

image

legacy.py
```diff
@@ -0,0 +1,22 @@
...
+	def __init__(self, name, **kwargs):
+		self.name = name
+		self.kwargs = kwargs
+
+
+LEGACY_PATH_MAP = {
+	'/index.php': Page('index'),
+	'/about.php': Page('about'),
+	'/trends.php': Page('report', report_id='state-of-the-web'),
+	'/interesting.php': Page('report', report_id='state-of-the-web', start='latest')
```
igrigorik: Should we also point http://httparchive.org/downloads.php to some docs on GitHub, or some such? We can drop the MySQL bits, but it'd be nice to preserve guidance on how to access raw archives and BigQuery data, wdyt?

main.py
```diff
@@ -160,7 +163,21 @@ def bad_request(e):
...
 @app.errorhandler(404)
 def page_not_found(e):
-	return render_template('error/404.html', error=e), 404
+	url = urlparse(request.url)
+	path = url.path
+	if legacy_util.should_redirect(path):
+		page = legacy_util.get_redirect_page(path)
+		redirect_url = url_for(page.name, **page.kwargs)
+		response = make_response(redirect(redirect_url, code=301))
+		# Set a cookie that expires 5 seconds after page load,
```
igrigorik: How confident are you that the page will load in under 5s? ;-)

templates/main.html
```diff
@@ -10,6 +10,15 @@
 {% endblock %}
 
 {% block content %}
+	{% if 'legacy_welcome' in request.cookies %}
+		<div class="alert alert-warning alert-dismissable" role="alert">
+			<strong>Welcome to the new HTTP Archive!</strong>
```
igrigorik: Link "new HTTP Archive" to the post as well?

legacy.py
rviscomi: Should we redirect it to https://beta.httparchive.org/faq#how-do-i-use-bigquery-to-write-custom-queries-over-the-data ?

main.py
rviscomi: Surprisingly no HTTP Archive data is available in CrUX for me to confirm, but I'm willing to accept the graceful degradation. :smile:

templates/main.html
rviscomi: Done

image

legacy.py
igrigorik: Yeah, that looks good.

rviscomi:
Added the redirect for /downloads.php. PTAL

PR #48, opened 2018-07-19
default CrUX maxDate to latest in GCS

rviscomi:
- new report config parameter `maxDateMetric`
- check param when setting up report JSON
- query GCS bucket to find latest date containing the metric
- cache latest date to prevent unnecessary lookups

We might want to consider a variation of this functionality for all reports, eg to omit dates from the Lighthouse report dropdowns when the metrics became unavailable in June.

Available for testing on staging: https://staging.httparchive.org/reports/chrome-ux-report

@igrigorik PTAL

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -563,6 +563,7 @@
 		],
 		"minDate": "2017_10_01",
 		"datePattern": ".*_01$",
+		"maxDateMetric": "cruxFcp",
 		"timeseries": {
 			"enabled": false
 		},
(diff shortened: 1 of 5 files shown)


rviscomi:
Right, it's cached but the cache is cleared after at most 3 hours along with all other report data. Not sure about other GAE refreshes.

igrigorik:
Ok. I assume the GCS query is not super expensive, so this should all be reasonable.

As an aside, any reason why can't or shouldn't be using GAE's memcache to persist this and similar data?

rviscomi:
memcache is a good idea and something I'd like to start using.

rviscomi:
This has been deployed live. Verify the changes at https://httparchive.org/reports/chrome-ux-report

igrigorik:
Works as advertised. Thanks Rick 👍

Issue #57, opened 2018-07-29
Relative percentage change sometimes reads "Infinity%".


malchata:
Currently, if a starting value in a dataset is `0`, the relative percentage change of a trend will display as "Infinity%" (because dividing by `0` in JS returns `Infinity`):

<img width="226" alt="screen shot 2018-07-29 at 1 04 34 pm" src="https://user-images.githubusercontent.com/1635238/43369235-1919dad6-9330-11e8-9cfc-636784519ee2.png">

The culprit [is the `getChange` function in `timeseries.js` not checking if the initial value is `0`](https://github.com/HTTPArchive/httparchive.org/blob/master/src/js/timeseries.js#L61).

*Proposed fix:* Either hide the relative change labels in this case, or just insert something like a dash or "n/a".

PR #58, opened 2018-07-29
Proposed fix for #57.

malchata:
This contains a proposed fix for #57 where the relative change labels can read "Infinity%" if the initial value is `0`. Instead, this new code finds the first non-zero occurrence in the data set's primary metric and uses that as the starting value in lieu of `Infinity`.

### DIFF
--- a/src/js/timeseries.js
+++ b/src/js/timeseries.js
@@ -54,7 +54,7 @@ function getSummary(data, options) {
 	const o = data[data.length - 1];
 	const summary = getPrimaryMetric(o, options);
 	const metric = new Metric(options, summary);
-	
+
 	return metric.toString();
 }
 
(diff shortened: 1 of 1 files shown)


src/js/timeseries.js
```diff
@@ -63,7 +63,16 @@ function getChange(data, options) {
...
+	let oldestIndex;
+
+	for (let i = 0; i < data.length; i++) {
+		if (getPrimaryMetric(data[i], options) > 0) {
+			oldestIndex = i;
+			break;
+		}
+	}
+
+	const oldest = getPrimaryMetric(data[oldestIndex], options);
```
rviscomi: Before this line can you add

```js
if (oldestIndex === undefined) {
  return;
}
```

If all values are 0 we won't have an index and it'll probably show `NaN` or something. This should hide the change entirely.

rviscomi:
Looks good! Just one comment about an edge case. Thanks for looking into it!

rviscomi:
Just realized I can commit my suggested change directly onto the branch. Merging!

rviscomi:
Fixes #57

Issue #40, opened 2018-04-30
Enable viewing reports through "lenses"


rviscomi:
**Goal:** Provide a mechanism for users to apply one or more "lenses" through which to view the HTTP Archive reports. These lenses effectively filter the set of websites to a subset that match some condition. For example, all websites that run on WordPress.

**Strategy:**

- [x] 1. write a query that generates a list of URLs/pages to act as the whitelist for the lens
- [x] 2. alter the report generation script to integrate that whitelist of pages with each metrics' timeseries and histogram queries
  - [x] 2a. optionally replace YYYY_MM_DD placeholders in whitelist query with report date
  - [x] 2b. save the lens results to an identifiable subdirectory on the gs:// CDN
- [x] 3. build a UI to enable the addition/removal of a lens
  - [x] 3a. provide a way to persist the selected lens across page views
  - [x] 3b. alter the report UI to make clear when a lens has been applied
  - [x] 3c. provide URL shortcuts (eg wordpress.httparchive.org or httparchive.org/wordpress)
- [x] 4. alter the data visualization scripts to pull data from the correct lens on the CDN

**Stretch goals:**

- [ ] 5. support multiple concurrent lenses for side-by-side comparison
  - [ ] 5a. in the lens selector UI
  - [ ] 5b. in the data visualization UI

rviscomi:
Prototyping this with a WordPress lens.

(1) is solved by querying the `httparchive.wappalyzer.wordpress` table, limited to the particular date/client of the report.

Made progress on (2) with [this commit](https://github.com/rviscomi/bigquery/commit/28e6e968ef200ea0735498c505c0e5f921792d76). It kinda works for some metrics. It's always joining using the `url` field, which doesn't do the expected behavior for metrics that query the requests table. For that we need to join on `request.page = detected.url`.

(2a) substitution is necessary. See script changes above.

(2b) is resolved by adding the query results to the GCS CDN in a new subdir under `/reports` that is the name of the lens. For example https://cdn.httparchive.org/reports/wordpress/bytesTotal.json is the timeseries of WordPress total bytes.

Skipping (3) for the prototype and manually adding the lens to the page via querystring param, eg `?lens=wordpress`.

[...]

PR #68, opened 2018-08-15
Lens UI

rviscomi:
Progress on https://github.com/HTTPArchive/httparchive.org/issues/40 UI to add/remove lenses.

Lenses are surfaced in the following UI:
- in the subdomain or querystring params of the URL (for permalinking)
- prefacing the report name on report pages (as FYI)
- in the permalink form on report pages (for adding/removing lenses)
- in individual chart titles (to disambiguate screenshots)

image

The field in the permalink form is a dropdown with an empty option (no lens), and an option for each lens name. Selecting an option will automatically update the permalink with the appropriate lens URL param. Clicking "Update" will reload the page with the lens enabled or disabled. I also added an "i" tooltip to describe what a lens is and hide it on mobile, where this is no mouseover.

It doesn't currently mess around with the subdomains. Stacking a `?lens=drupal` on top of `wordpress.httparchive.org` in this UI will set the lens to Drupal. Ideally it would drop the URL param and change the host to `drupal.httparchive.org`, but it seems ok for now. I left a TODO in the code.

The ordering of the lenses is non-deterministic. Also left a TODO to sort them.

### DIFF
--- a/src/js/report.js
+++ b/src/js/report.js
@@ -10,9 +10,11 @@ class Report {
 		this.report = report;
 		this.viz = viz;
 		this.baseUrl = report.url;
+		this.lens = report.lens && report.lens.id;
 		this.startDate = report.startDate;
 		this.endDate = report.endDate;
 
+		this.bindChangeListener('lens');
 		this.bindChangeListener('startDate');
 		this.bindChangeListener('endDate');
 		this.bindUpdateListener();
@@ -80,12 +82,21 @@ class Report {
 	}
 
 	updatePermalink() {
+		const url = new URL(this.baseUrl);
+		const lens = this.lens;
+
+		// TODO: Change subdomain.
+		if (lens) {
+			url.searchParams.set('lens', lens);
+		} else {
+			url.searchParams.delete('lens');
+		}
+
 		if (this.isOneYearAgo(this.startDate) && this.isLatest(this.endDate)) {
-			this.permalink.value = this.baseUrl;
+			this.permalink.value = url.toString();
 			return;
 		}
 
-		const url = new URL(this.baseUrl);
 		const start = this.getDateUrlAlias(this.startDate);
 		const end = this.getDateUrlAlias(this.endDate);
 
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -287,6 +287,14 @@ input {
 	border-bottom-style: solid;
 }
 
+.help-tooltip {
+	opacity: 0.7;
+	transition: all 0.2s;
+}
+.help-tooltip:hover {
+	opacity: 1.0;
+}
+
 
 @media (max-width: 992px) {
 	footer {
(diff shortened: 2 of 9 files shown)


rviscomi:
@igrigorik you can demo this functionality on https://staging.httparchive.org/

igrigorik:
Sooo awesome! Nice work Rick. 

Some of the Magento stats are slightly frightening. Need to dig in to understand what's going on there, but that's not a comment about the reports, but the underlying causes behind those stats.

One small gotcha I ran into: https://staging.httparchive.org/reports/loading-speed?lens=wordpress&start=2017_08_01&end=latest#bootupJs — doesn't load for WP, but works for others?

src/js/report.js
```diff
@@ -80,12 +82,21 @@ class Report {
 	}
 
 	updatePermalink() {
+		const url = new URL(this.baseUrl);
+		const lens = this.lens;
+
+		// TODO: Change subdomain.
```
igrigorik: Would we update the subdomain *and* add the lens query param, or just the former?

static/css/styles.css
```diff
@@ -287,6 +287,14 @@ input {
 	border-bottom-style: solid;
 }
 
+.help-tooltip {
```
igrigorik: Hmm, the tooltip is not coming up for me when I hover. Also, what happens for mobile clients? 

Instead of a tooltip, perhaps we should create a FAQ entry that explains what lenses do, and how the url classification is done (via Wappalyzer)?

src/js/report.js
rviscomi: We would update the subdomain instead of the param.

static/css/styles.css
rviscomi: +1 to FAQ, I'll have the icon link to that instead of a tooltip

src/js/report.js
igrigorik: What's the TODO there, is it setting up DNS, or is it more involved? Could we setup a wildcard such that we don't have to manually add DNS records each time?

rviscomi: There are two steps needed for adding a subdomain:

- register it on Google Domains
- register it with the GCP load balancer, which also enables TLS

I've already done these steps for the new [Drupal](drupal.httparchive.org) and [Magento](magento.httparchive.org) lenses. I'll look into ways to use wildcards to avoid the need for these steps for future lenses.

The TODO here is to parse the permalink URL and add/remove/replace the subdomain accordingly rather than relying on the `lens` URL param. The nice thing about the param is that it works out of the gate for all new lenses and there's already a supported [`searchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) API for add/remove/replace actions.

rviscomi:
> One small gotcha I ran into: https://staging.httparchive.org/reports/loading-speed?lens=wordpress&start=2017_08_01&end=latest#bootupJs — doesn't load for WP, but works for others?

Fixed thanks!

static/css/styles.css
rviscomi: Added PTAL

rviscomi:
🔔 ping :)

static/css/styles.css
igrigorik: 👍

PR #81, opened 2018-09-07
Make the charts taller

paulirish:
This probably should have started with some discussion rather than the PR, but it was so easy!

Whenever I'm looking at HA charts, the lack of vertical height makes it harder for me to see the changes over time. This PR just makes the charts 50% taller.

Before and after:
image

wdyt?

### DIFF
--- a/static/css/report.css
+++ b/static/css/report.css
@@ -143,7 +143,7 @@ th {
 
 .chart {
 	min-width: 310px;
-	height: 400px;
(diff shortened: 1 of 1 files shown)


rviscomi:
Makes total sense. Thanks!

igrigorik:
Late to the party but 👍👍👍

rviscomi:
This change is live at httparchive.org

PR #84, opened 2018-09-08
minor layout fix

wellingguzman:
Removing the `margin` from the columns make each column centered to the view on smaller viewport. It also makes use of the `float` instead of changing the display to `inline-block`, so it keeps the Bootstrap layout system.

Example 1:
kapture 2018-09-08 at 12 34 38

Example 2:
kapture 2018-09-08 at 12 22 24

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -383,11 +383,11 @@ input {
 }
 
 @media (min-width: 768px) {
-	.container-flex {
+	.flex-container {
 		display: flex;
 	}
 
-	.container-flex.align-middle {
+	.flex-align-middle {
 		align-items: center;
 	}
 }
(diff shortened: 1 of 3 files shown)


rviscomi:
There are a few subtle side effects of this change. For example, the graphic on the home page loses its vertical alignment.

Before:
image

After:
image

Images in Bootstrap inherit a `vertical-align: middle;` style which relies on its parent container being an inline-block.

@WellingGuzman could you take a look at the image alignment issue? Maybe the fix is to be more explicit with the alignment (eg using flexbox). Any other ideas?

rviscomi:
There are other parts of the page that need to be explicitly aligned.  Can you go through the site to check for others? Here are two I found on the home page:

Before:
image

After:
image

Before: 
image

After:
image

WellingGuzman:
There's may be some more places where this is happening, I will give a another round to review the two places. I am unable to build the site, I am actually testing this on the live site using the devtools.

I have updated the footer menu and the big query container to have its data centered using flexbox.

rviscomi:
Hey @WellingGuzman 👋 any update on this?

Issue #88, opened 2018-09-19
ModuleNotFoundError: No module named 'urlparse'


WellingGuzman:
```
$ python --version
Python 3.6.4
```

Running `python main.py` returns the following error:

```
$ python main.py
Traceback (most recent call last):
  File "main.py", line 19, in <module>
    from urlparse import urlparse
ModuleNotFoundError: No module named 'urlparse'
```

rviscomi:
We should enforce a Python version somewhere.

PR #89, opened 2018-09-19
Fix importing urlparse in python 3

WellingGuzman:
Running the server using Python 3 throws an error as it can't find the `urlparse` module.

I came across the same issue here: https://github.com/heroku/kafka-helper/issues/6 (with the solution)

Is there a python version requirement for the project?

Ref: #88

### DIFF
--- a/main.py
+++ b/main.py
@@ -16,7 +16,10 @@
 import logging
 import re
 from time import time
-from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


rviscomi:
Ah. I'm running 2.7. This fix LGTM.

PR #91, opened 2018-09-21
Pull changelog.json from cdn.rawgit

paulirish:
We noticed some  503's that caused the graphs to not load. Hypothetically this avoids the 503s. thx https://rawgit.com/

timeseries.js should probably also assume this fetch _could_ fail and still render the graphs anyway.

### DIFF
--- a/csp.py
+++ b/csp.py
@@ -21,7 +21,7 @@ csp = {
 		'\'self\'',
 		'cdn.httparchive.org',
 		'discuss.httparchive.org',
-		'raw.githubusercontent.com',
+		'cdn.rawgit.com',
 		'www.webpagetest.org',
 		'www.google-analytics.com',
 		'stats.g.doubleclick.net'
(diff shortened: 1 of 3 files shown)


WellingGuzman:
@paulirish if it matters, it's also being used here:

- https://github.com/HTTPArchive/httparchive.org/blob/43a26a77d84fb8547564518a8469be69ed852cf1/src/js/utils.js#L62
- https://github.com/HTTPArchive/httparchive.org/blob/c7f74944d777739478f85ae0875a882e059e8643/csp.py#L24

Probably it's worth to change it in these places too.

paulirish:
@WellingGuzman 	hey man! 👋   thx for the heads up. did those too.

rviscomi:
Thanks for spotting and fixing this @paulirish! LGTM :+1:

rviscomi:
Turns out rawgit is shutting down soon so we'll need to replace this solution 😞

PR #103, opened 2018-10-29
Improve Tablet Navigation

stephanmax:
In the end, I went for a simpler way than initially proposed. Open-on-click navigational menus pose their own challenges (Do they have to close on click-away?), so I left them as open-on-hover/focus and just removed the links of the dropdown targets.

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -43,7 +43,6 @@ nav li {
 }
 nav li a,
 nav li .dropdown-target {
-	cursor: pointer;
 	display: inherit;
 	padding: 12px 20px;
 }
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -43,6 +43,7 @@ nav li {
 }
 nav li a,
 nav li .dropdown-target {
+	cursor: pointer;
 	display: inherit;
 	padding: 12px 20px;
 }
--- a/templates/main.html
+++ b/templates/main.html
@@ -36,12 +36,13 @@
 				<nav class="hidden-xs pull-right text-right">
 					<ul>
 						<li>
-							<a class="dropdown-target" aria-haspopup="true" href="{{ url_for('reports') }}">
+							<a class="dropdown-target" aria-haspopup="true">
 								Reports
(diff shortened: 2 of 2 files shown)


static/css/styles.css
```diff
@@ -43,6 +43,7 @@ nav li {
 }
 nav li a,
 nav li .dropdown-target {
+	cursor: pointer;
```
rviscomi: What's the reasoning behind adding this style? IMO it makes the target behave more like a clickable link.

templates/main.html
```diff
@@ -36,12 +36,13 @@
 				<nav class="hidden-xs pull-right text-right">
 					<ul>
 						<li>
-							<a class="dropdown-target" aria-haspopup="true" href="{{ url_for('reports') }}">
+							<a class="dropdown-target" aria-haspopup="true">
```
rviscomi: Could we use a different element here rather than `<a>` if we're removing the link functionality from it? Eg maybe `<span>`.

rviscomi:
A couple of comments. Thanks for working on this @stephanmax!

static/css/styles.css
stephanmax: Good point! My thinking was that the target still behaves like a menu button and should get the pointer cursor rather than the generic text selection cursor to depict it as clickable.

templates/main.html
stephanmax: You are right. Not per se against `span`, but do you think `button` would be more semantic. That would also "solve" the comment above.

rviscomi: It's a hard call. `button`s typically change the state of the application. Is the dropdown menu open/close changing states? Maybe? As long as we persist the `aria-haspopup` attribute I don't think it matters much what element we use. But I do think leaving it as `a` has some unintended side-effects, like the underline on hover, and is even less semantic than span/button. I'll leave it up to you :)

static/css/styles.css
rviscomi: The target doesn't do anything on click, so a pointer cursor sends a mixed message. Besides, by the time the user hovers over it to change the cursor, the popup is already open.

stephanmax:
Changed accordingly.

rviscomi:
LGTM thanks again!

Issue #87, opened 2018-09-19
Allow local building without gcloud


rviscomi:
Authenticating with Google Cloud is necessary to call the storage APIs that tell us which dates are available for reports. When building the website locally, it should be possible to skip this requirement and get a truly static/deterministic environment. This can be with a non-default build flag or something.

voidpirate:
Implemented in https://github.com/HTTPArchive/httparchive.org/pull/104, should be safe to close out this issue :)

PR #104, opened 2018-10-29
Add exception handling for failed auth to GCS

voidpirate:
This should fix issue [Allow local building without gcloud](https://github.com/HTTPArchive/httparchive.org/issues/87) when `GOOGLE_APPLICATION_CREDENTIALS` are not configured locally. Instead of adding a build flag, this just catches the exception raised when authentication fails and sets a flag to not load the gcs bucket.

I added a custom logger to `dates.py` so when developing you can easily see when the site isn't loading dates from GCS.

```
dates.py - WARNING - Unable to authenticate to Google Cloud Storage.
dates.py - WARNING - Google Cloud Storage disabled, skipping get_dates()
```

Additionally, I couldn't decide on an elegant way to fail the route `/reports/<report_id>` since it requires dates to be available, so I log a message to console (`WARNING:root:Unable to load dates from GCS`) and aborts with a 500, open to suggestions on this.

This will probably need to be tested in GAE flex to make sure I didn't break anything :)

### DIFF
--- a/dates.py
+++ b/dates.py
@@ -1,10 +1,62 @@
 import re
+import logging
+
 from google.cloud import storage
+from google.auth.exceptions import DefaultCredentialsError
+
+# Create/configure stream handler
+sh = logging.StreamHandler()
+fmt = logging.Formatter('%(name)s.py - %(levelname)s - %(message)s')
+sh.setFormatter(fmt)
+
+# Register stream handler
+logger = logging.getLogger(__name__)
+logger.addHandler(sh)
 
 GCS_BUCKET = 'httparchive'
 
-def get_dates():
+# When running the site locally and unable to authenticate with GCS bypass
+# loading dates from GCS.
+LOAD_DATES_FROM_GCS = True
+try:
 	gcs = storage.Client()
+except DefaultCredentialsError:
+	logger.warning('Unable to authenticate to Google Cloud Storage.')
+	LOAD_DATES_FROM_GCS = False
+
+	def mock_get_dates(start_year, end_year, today, month_delta=0):
+		''' Generate mock dates to the format the site frontend expects them'''
+		year = start_year
+		months = []
+		for month in range(1, ((end_year - start_year) * 12) + (month_delta + 1)):
+			current_month_in_year = month % 12
+			if current_month_in_year == 0:
+				current_month_in_year = 12
+
+			if current_month_in_year < 10:
+				current_month_in_year = "0{}".format(current_month_in_year)
+
+			months.append('{}_{}_01'.format(year, current_month_in_year))
+			if year == end_year and current_month_in_year == month_delta and today < 15:
+				continue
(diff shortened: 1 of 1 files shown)


rviscomi:
> Additionally, I couldn't decide on an elegant way to fail the route /reports/<report_id> since it requires dates to be available, so I log a message to console (WARNING:root:Unable to load dates from GCS) and aborts with a 500, open to suggestions on this.

Could we "guess" what the dates should look like? Or maybe use a static list of known good dates? I think it'd be nice to fail in a way that still enables the page to be usable for developers when testing.

voidpirate:
I've added helper methods that will mock the dates if GCS isn't available. Site locally now looks to be running similar to the live site.

dates.py
```diff
@@ -1,10 +1,59 @@
...
+from google.auth.exceptions import DefaultCredentialsError
+
+# Create/configure stream handler
+sh = logging.StreamHandler()
+fmt = logging.Formatter('%(name)s.py - %(levelname)s - %(message)s')
+sh.setFormatter(fmt)
+
+# Register stream handler
+logger = logging.getLogger(__name__)
+logger.addHandler(sh)
```
rviscomi: This logging boilerplate looks like it would be really useful for other files. Note to self that this would be useful in its own util. (no changes needed in this PR)

dates.py
```diff
@@ -1,10 +1,59 @@
...
+		months = []
+		for month in range(1, ((end_year - start_year) * 12) + month_delta):
+			current_month_in_year = month % 12
+			if current_month_in_year == 0:
+				current_month_in_year = 12
+
+			if current_month_in_year < 10:
+				current_month_in_year = "0{}".format(current_month_in_year)
+
+			months.append('{}_{}_{}'.format(current_year, current_month_in_year, '01'))
```
rviscomi: Out of curiosity, is there an advantage to using `format` as opposed to string formatting, eg:

```py
months.append(`%s_%s_01` % (current_year, current_month_in_year))
```

dates.py
```diff
@@ -1,10 +1,59 @@
...
+			months.append('{}_{}_{}'.format(current_year, current_month_in_year, '15'))
+
+			if month % 12 == 0:
+				current_year = current_year + 1
```
rviscomi: nit: can we name this `year`?

main.py
```diff
@@ -102,9 +102,6 @@ def report(report_id):
 		abort(404)
 
 	dates = report_util.get_dates()
-	if not dates:
```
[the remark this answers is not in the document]
voidpirate: Yeah good call, added it back.

dates.py
voidpirate: Just a personal preference as I find it more expressive in some cases - I could go either way with this :)

voidpirate:
Should be good for another review :)

Issue #109, opened 2018-11-08
[Loading Speed] Metrics rename


addyosmani:
I've been asking touchpoints leveraging Lighthouse for metrics data to rename TTCI and TTFI now that we've done so in Lighthouse:

Time To First Interactive is now First CPU Idle
Time to Consistently Interactive is now just Time to Interactive

(SpeedCurve also just made this change)

rviscomi:
This is an easy fix for anyone looking to contribute their first PR to the project. All that's needed is to change the `name` property of these metric configs:

https://github.com/HTTPArchive/httparchive.org/blob/dddfe085b43894b3a506099608afe1f9e5d5bed3/config/reports.json#L301-L322

Any takers?

denar90:
Maybe we can create some global Progressive Web Metrics (PWM) config (repo) to use across all projects, as we do in pwmetrics - https://github.com/paulirish/pwmetrics/blob/master/lib/metrics/metrics.ts

So tip for other will be, just use this file and that's it. Updating it will notify that "spec" of PWM changed, then authors of libs should bump version.

The problem here, how to be with configs in json format, maybe people should migrate to exporting object which will be mixed with PWM object?

denar90:
> Any takers?

I'm in :)

rviscomi:
Fixed! Thanks again @denar90!

![image](https://user-images.githubusercontent.com/1120896/48292332-e6c0b500-e447-11e8-830e-9c9e129d1bb2.png)

PR #110, opened 2018-11-09
WIP: Rename interactive metrics

denar90:
Solves #109

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -298,7 +298,7 @@
 				]
 			}
 		},
-		"tti": {
+		"ttci": {
 			"name": "Time to Interactive",
 			"type": "seconds",
 			"singular": "second",
@@ -307,7 +307,7 @@
 				"minDate": "2017_06_01"
 			}
 		},
-		"firstCPUIdle": {
+		"ttfi": {
 			"name": "First CPU Idle",
 			"type": "seconds",
 			"singular": "second",
@@ -654,8 +654,8 @@
 		"summary": "Web performance can directly impact business metrics like conversion and user happiness. This report analyzes various performance metrics in the lifecycle of a loading page including those used by many modern progressive web apps.",
 		"metrics": [
 			"fcp",
-			"tti",
-			"firstCPUIdle",
+			"ttci",
+			"ttfi",
 			"dcl",
 			"ol",
 			"bootupJs"
--- a/config/reports.json
+++ b/config/reports.json
@@ -298,20 +298,20 @@
 				]
 			}
 		},
-		"ttci": {
-			"name": "Time to Consistently Interactive",
+		"tti": {
+			"name": "Time to Interactive",
 			"type": "seconds",
 			"singular": "second",
-			"description": "The number of seconds from the time the navigation started until the page is fully interactive and the network is idle. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/consistently-interactive) as is only available in mobile tests.",
+			"description": "The number of seconds from the time the navigation started until the page is fully interactive and the network is idle. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/time-to-interactive) as is only available in mobile tests.",
 			"histogram": {
 				"minDate": "2017_06_01"
 			}
 		},
-		"ttfi": {
-			"name": "Time to First Interactive",
+		"firstCPUIdle": {
+			"name": "First CPU Idle",
 			"type": "seconds",
 			"singular": "second",
-			"description": "The number of seconds from the time the navigation started until the layout is stabilized, web fonts are visible, and the page is responsive to user input. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/consistently-interactive) as is only available in mobile tests.",
+			"description": "The number of seconds from the time the navigation started until the layout is stabilized, web fonts are visible, and the page is responsive to user input. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/time-to-interactive) as is only available in mobile tests.",
 			"histogram": {
 				"minDate": "2017_06_01"
 			},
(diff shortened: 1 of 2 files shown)


config/reports.json
```diff
@@ -298,20 +298,20 @@
...
 			}
 		},
-		"ttci": {
-			"name": "Time to Consistently Interactive",
+		"tti": {
+			"name": "Time to Interactive",
 			"type": "seconds",
 			"singular": "second",
-			"description": "The number of seconds from the time the navigation started until the page is fully interactive and the network is idle. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/consistently-interactive) as is only available in mobile tests.",
+			"description": "The number of seconds from the time the navigation started until the page is fully interactive and the network is idle. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/time-to-interactive) as is only available in mobile tests.",
```
denar90: wdyt about replacing https://developers.google.com/web/tools/lighthouse/audits/time-to-interactive by https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c ?

denar90:
It can be merged only after renaming (I suppose) [ttci.sql](https://github.com/HTTPArchive/bigquery/blob/0691ea5ef318f0589172f3dcb8ed612be3f3fe97/sql/histograms/ttci.sql) and [ttfi.sql](https://github.com/HTTPArchive/bigquery/blob/0691ea5ef318f0589172f3dcb8ed612be3f3fe97/sql/histograms/ttfi.sql).

config/reports.json
rviscomi: Oh nice, didn't know about that doc. Can we link to both from each metric's description? I suggest we just append a new sentence, "Read more."

config/reports.json
```diff
@@ -298,20 +298,20 @@
 				]
 			}
 		},
-		"ttci": {
```
rviscomi: As you pointed out, changing the metric ID here also requires renaming the SQL files in the bigquery repo. It also requires renaming the JSON files on the CDN. For example:

- https://cdn.httparchive.org/reports/ttci.json (timeseries)
- https://cdn.httparchive.org/reports/wordpress/ttci.json (lens timeseries)
- https://cdn.httparchive.org/reports/2018_10_15/ttci.json (histogram)
- https://cdn.httparchive.org/reports/wordpress/2018_10_15/ttci.json (lens histogram)

All of those for both metrics, for all 3 lenses, and for all histograms since 2017_06_01.

That's a lot of files. It's doable with a script, but not sure if it's worth the effort. Alternatively, I propose we keep the metric IDs as-is and simply update the `name` property. That would eliminate the need for renaming the JSON and SQL files. How does that sound?

denar90: I'm totally ok with that, one concern I have is related to understanding and reading the code. Newcomers might be confused by not consistent key names.
But we can do renamings later...

config/reports.json
```diff
@@ -299,19 +299,19 @@
 			}
 		},
 		"ttci": {
-			"name": "Time to Consistently Interactive",
+			"name": "Time to Interactive",
 			"type": "seconds",
 			"singular": "second",
-			"description": "The number of seconds from the time the navigation started until the page is fully interactive and the network is idle. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/consistently-interactive) as is only available in mobile tests.",
+			"description": "The number of seconds from the time the navigation started until the last long task happened before 5 seconds window where network is mostly quiet. [Read more](https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#heading=h.iqlwzaf6lqrh).",
```
denar90: There no clear definition in spec file, so I improvised a bit here.

config/reports.json
```diff
@@ -299,19 +299,19 @@
...
 				"minDate": "2017_06_01"
 			}
 		},
 		"ttfi": {
-			"name": "Time to First Interactive",
+			"name": "First CPU Idle",
 			"type": "seconds",
 			"singular": "second",
-			"description": "The number of seconds from the time the navigation started until the layout is stabilized, web fonts are visible, and the page is responsive to user input. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/consistently-interactive) as is only available in mobile tests.",
+			"description": "The number of seconds from the time the navigation started until the minimally page interactive. [Read more](https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#heading=h.d06i5ycfxaow).",
```
denar90: It's kinda mix of prev description and https://developers.google.com/web/tools/lighthouse/audits/first-cpu-idle, because of not clear definition to copy-paste.

config/reports.json
rviscomi: SGTM 👍

config/reports.json
rviscomi:
```suggestion
			"description": "The number of seconds from the time the navigation started until the CPU had at least 5 seconds of quiescence. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/time-to-interactive) as is only available in mobile tests. [Read more](https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#heading=h.iqlwzaf6lqrh).",
```

config/reports.json
rviscomi:
```suggestion
			"description": "The number of seconds from the time the navigation started until the page is minimally interactive. This metric comes from [Lighthouse](https://developers.google.com/web/tools/lighthouse/audits/first-cpu-idle) as is only available in mobile tests. [Read more](https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#heading=h.d06i5ycfxaow).",
```

denar90:
💃

addyosmani:
👍

PR #117, opened 2019-01-02
improve color contrast

### LINKED ISSUE #115: Fix color contrast of navigation links on hover
I noticed that the color contrast on `nav li:hover` is way to low (https://contrast-ratio.com/#white-on-%23bcced1). There might be other places where this is an issue. 

I could help with that, but I did not create a pull request since design/branding changes are always controversial.

xi:
fixes #115 

Still far from perfect, but hopefully much better than before.

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -722,7 +722,7 @@
 			"bgcolor": "#f7df1e",
 			"primary": {
 				"text": "JS",
-				"color": "#555"
+				"color": "#444"
 			},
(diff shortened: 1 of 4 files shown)


rviscomi:
Thanks @xi! Could you share a couple of before/after screenshots?

xi:
Except for the menu the changes are subtle.

Before:

before

After:

after

Diff:

diff

PR #134, opened 2019-03-21
fix(about.html): add target="_blank" and rel="nofollow" - Closes #132

### LINKED ISSUE #132: Use rel=nofollow on sponsor links
To avoid the impression that sponsors get SEO benefits for supporting the HTTP Archive, we should apply `rel="nofollow"` to all outbound sponsor links.

tjmonsi:
This is to avoid impression that sponsors get SEO benefits for
supporting httparchive

Closes #132

### DIFF
--- a/templates/about.html
+++ b/templates/about.html
@@ -56,34 +56,58 @@
 
 			<p>
 				The HTTP Archive is possible through the support of these sponsors:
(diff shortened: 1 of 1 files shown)


rviscomi:
Thanks for fixing this @tjmonsi!

Issue #165, opened 2019-06-07
CSP blocking discourse avatars


anoblet:
I'm not able to see profile images on the home page. Here's a screenshot:

![CSP](https://user-images.githubusercontent.com/7674171/59076927-73c8d100-88a6-11e9-8d3f-1aff271a81e8.png)

rviscomi:
Good catch! Looks like the discourse avatar domain changed and we need to add it to the CSP img policy. The domain to add is `sjc3.discourse-cdn.com`.

@anoblet could you file a PR to add the domain to the [img policy](https://github.com/HTTPArchive/httparchive.org/blob/master/csp.py#L30)?

anoblet:
Sure, I'll take care of it today.

rviscomi:
New problem. Now it seems the avatars are 404ing. I don't think think the CSP change caused it, but it seems to have uncovered it. @anoblet the solution here isn't very clear, do you want to give it a go anyway?

![image](https://user-images.githubusercontent.com/1120896/59131293-01a4ca80-8940-11e9-8e78-bee5acd934d5.png)

anoblet:
Sure

anoblet:
There seems to be an issue with this line: https://github.com/HTTPArchive/httparchive.org/blob/d4faadadfcd37635ad6d2687ac4e90e74a6fd918/src/js/discussion.js#L102

As this URL works: https://sjc3.discourse-cdn.com/standard17/user_avatar/discuss.httparchive.org/rviscomi/75/835_2.png

rviscomi:
Yeah good catch! Could we try updating the size replacement to 50 or 75?

This URL also works https://sjc3.discourse-cdn.com/standard17/user_avatar/discuss.httparchive.org/rviscomi/50/835_2.png

rviscomi:
Fixed by #167. Tested locally and it works, but deployed and still seeing broken images. Redeploying to see if it was a fluke and leaving this issue open until I can verify it.

Looking at this widget, I see another bug, which is that the same post may be listed twice. Filed #168. @anoblet that one is more about JS coding if you'd like to take it.

anoblet:
I'm working on setting up the stack on my computer. As of right now I see it working on httparchive.org.

rviscomi:
Working for me too now, I think something got stuck in my cache. We can close this now.

Thanks again for your help finding and fixing this issue!

PR #166, opened 2019-06-07
Add discourse cdn to csp policy

anoblet:
Fixes: https://github.com/HTTPArchive/httparchive.org/issues/165


rviscomi:
Thanks for fixing this @anoblet!

Issue #168, opened 2019-06-07
Dedupe featured discussions


rviscomi:
The home page pulls in lists of 2 recent and 3 top discussions on the [forum](https://discuss.httparchive.org). Sometimes a recent discussion is also a top discussion, like the Almanac post below:

![image](https://user-images.githubusercontent.com/1120896/59134831-aaf0be00-894a-11e9-8251-37d75ece6898.png)

If a top post is also a recent post, include it in the list of 2 recent posts but choose a third post for the list of top posts. Keep choosing new top posts until we get one that isn't also recent.

rviscomi:
I think the fix here is to create a Set of the 2 most recent post IDs and filter them out of the list of top posts before choosing the top 3.

rviscomi:
#170 was merged and deployed to the [staging server](https://staging.httparchive.org/). However it's still showing duplicate posts in the top/latest lists. Reopening this to investigate.

rviscomi:
As soon as I wrote that I tested it one last time and it must have been a caching issue. It's working for me now on staging. I'll deploy to the live site and re-close this issue 😊

PR #170, opened 2019-06-08
Fix duplicate topics

anoblet:
Fixes: https://github.com/HTTPArchive/httparchive.org/issues/168

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -12,17 +12,12 @@ function getDiscussTopics() {
 		const topics = r.topic_list.topics.slice(0, 2);
 		topics.forEach(topic => latestTopicIds.add(topic.id));
 		drawTopics(topics, r.users);
-	}).then(_ =>
-		fetch(`${Discussion.ORIGIN}/top.json`).then(r => r.json()).then(r => {
-			const rTopics = r.topic_list.topics;
-			let topics = [];
-			for (var i = 0; i < rTopics.length; i++) {
-				const topic = rTopics[i];
-				if(!idList.includes(topic.id)) topics.push(topic);
-				if(topics.length > 2) break;
-			}
-			drawTopics(topics, r.users);
-		})
+	}).then(_ => {
+			fetch(`${Discussion.ORIGIN}/top.json`).then(r => r.json()).then(r => {
+				const topics = r.topic_list.topics.filter(topic => !latestTopicIds.has(topic.id)).slice(0, 3);
+				drawTopics(topics, r.users);
+			})
+		}
 	).then(_ => section.classList.remove('hidden'))
 }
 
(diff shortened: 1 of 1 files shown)


anoblet:
This pulls 3 popular topics, then iterates the latest topics until it can find two which weren't part of the popular set.

src/js/index.js
```diff
@@ -7,25 +7,31 @@ function getDiscussTopics() {
...
-		const allUsers = new Map(r.users.map(user => [user.id, user]));
-
-		if (topics && topics.length && allUsers.size) {
-			drawTopics(topics, allUsers);
-		}
-	}).then(_ => {
-		return fetch(`${Discussion.ORIGIN}/top.json`);
-	}).then(r => r.json()).then(r => {
+	let idList = [];
+	fetch(`${Discussion.ORIGIN}/top.json`).then(r => r.json()).then(r => {
```
rviscomi: The order of top/latest was switched. We actually want to filter recent posts out of top, not the other way around.

anoblet: I realized that after I had pushed it :) How does this look?

anoblet: I fixed the top/latest priority issue and did a little bit of cleanup.

src/js/index.js
```diff
@@ -7,36 +7,37 @@ function getDiscussTopics() {
 	if (!section || !discussions) {
 		return;
 	}
+	let idList = [];
```
rviscomi: Small optimization, but could you use a Set instead? Also rename the variable to something more descriptive:

```suggestion
	let latestTopicIds = new Set();
```

src/js/index.js
```diff
@@ -7,36 +7,37 @@ function getDiscussTopics() {
...
-		const topics = r.topic_list.topics.slice(0, 3);
-		const allUsers = new Map(r.users.map(user => [user.id, user]));
-
-		if (topics && topics.length && allUsers.size) {
-			drawTopics(topics, allUsers);
-		}
-	}).then(_ => {
-		section.classList.remove('hidden');
-	});
+		topics.map((topic) => idList.push(topic.id));
```
rviscomi: `map` is best for mutating the array. `forEach` is more appropriate here.

```suggestion
		topics.forEach(topic => latestTopicIds.add(topic.id));
```

src/js/index.js
```diff
@@ -7,36 +7,37 @@ function getDiscussTopics() {
...
+		drawTopics(topics, r.users);
+	}).then(_ =>
+		fetch(`${Discussion.ORIGIN}/top.json`).then(r => r.json()).then(r => {
```
rviscomi: Could you use the multiline `_ => { return fetch(); }` style here to avoid chaining `then`s on the same line?

src/js/index.js
```diff
@@ -7,36 +7,37 @@ function getDiscussTopics() {
...
+			const rTopics = r.topic_list.topics;
+			let topics = [];
+			for (var i = 0; i < rTopics.length; i++) {
```
rviscomi: I'd keep the `slice` from before, but just add a filter before it:

```suggestion
			const topics = r.topic_list.topics.filter(topic => !latestTopicIds.has(topic.id)).slice(0, 3);
```

rviscomi:
Thanks @anoblet! 🎉

PR #182, opened 2019-10-05
Specify page weight may be compressed

### LINKED ISSUE #137: Specify "may be compressed" in Page Weight report
Per [this discussion](https://discuss.httparchive.org/t/css-bytes-reports-gzip-vs-raw-sizes/1618/4?u=rviscomi) on the forum, it's not immediately clear whether the download size represents compressed or uncompressed bytes. This is actually question that comes up often.

We do say that the size represents the transfer size, which implies that the bytes may be compressed, but it's not immediately obvious what that term means. Per the suggestion in the thread, let's add a line to the report description that the weights shown may include compressed resources:

> This report tracks the size and quantity of many popular web page resources. Sizes represent the number of bytes sent over the network, which may be compressed.

The report description is defined in the config file:

https://github.com/HTTPArchive/httparchive.org/blob/5ef26fdc0324e1cb942927c1080c0328ed2ba076/config/reports.json#L796

wellingguzman:
Closes #137

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -793,7 +793,7 @@
 	},
 	"page-weight": {
 		"name": "Page Weight",
-		"summary": "This report tracks the size and quantity of many popular web page resources, which may be compressed.",
+		"summary": "This report tracks the size and quantity of many popular web page resources. Sizes represent the number of bytes sent over the network, which may be compressed.",
 		"metrics": [
 			"bytesTotal",
 			"reqTotal",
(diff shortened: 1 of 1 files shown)


config/reports.json
```diff
@@ -793,7 +793,7 @@
 	},
 	"page-weight": {
 		"name": "Page Weight",
-		"summary": "This report tracks the size and quantity of many popular web page resources.",
+		"summary": "This report tracks the size and quantity of many popular web page resources, which may be compressed.",
```
rviscomi:
```suggestion
		"summary": "This report tracks the size and quantity of many popular web page resources. Sizes represent the number of bytes sent over the network, which may be compressed.",
```

PR #206, opened 2020-06-25
Add Project Fugu report

tomayac:
This PR adds a new top-level report for Project Fugu APIs. It is sorted by the level of maturity of the APIs, starting from (🐡 Launched), over to (🧪 Origin trial), and ending with (🚩 Behind flag).

Some of the reports show no data, since the number of pages that implement these APIs currently is still low. I round to 5 digits in the queries, which, if my math is right and assuming the current size of 5,350,650 URLs of the mobile archive, that there must be at least 0.00001 * 5,350,650 ≈ 54 pages using an API in order for the report to show data.

Please review this jointly with https://github.com/HTTPArchive/bigquery/pull/84.

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -818,7 +818,7 @@
 			}
 		},
 		"shapeDetectionFace": {
-			"name": "(🏁 Behind flag) Shape Detection (Face)",
+			"name": "(🚩 Behind flag) Shape Detection (Face)",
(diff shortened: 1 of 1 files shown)


rviscomi:
Staged here https://staging.httparchive.org/reports/project-fugu

Issue #221, opened 2020-09-18
Fugu report has many failing queries


tomayac:
It seems like something broke with the latest data set release. A lot of queries in the [Fugu report](https://httparchive.org/reports/project-fugu) don't seem to return any data (for example [`idleDetection.json`](https://cdn.httparchive.org/reports/idleDetection.json)). @rviscomi, are you aware of any recent changes that could have caused this? 

<img width="766" alt="Screen Shot 2020-09-18 at 12 17 50" src="https://user-images.githubusercontent.com/145676/93586715-fd6fb580-f9a8-11ea-95b2-b7f7207a2c88.png">

rviscomi:
Thanks for catching this! I'll investigate.

rviscomi:
@tomayac the `idleDetection` query doesn't return any results. Adoption looks ~0% on [Chrome Status](https://www.chromestatus.com/metrics/feature/timeline/popularity/2834).

Do you expect these features to have data? If not, is the bug that the charts should still render "0%" when there is no data?

tomayac:
Before the charts displayed and there were no errors on the console.

tomayac:
I happen to have a static dump from the previous state: https://tomayac.github.io/temp/.

rviscomi:
Is it possible that the queries changed? The temp page is showing values of `0`s while the live page is getting no values at all (empty array).

tomayac:
I don't think anything changed.

rviscomi:
CORS is blocking the JSON data from loading on your temp page. Any idea where the data is coming from in those charts?
![image](https://user-images.githubusercontent.com/1120896/93626980-42e4b080-f9b2-11ea-9ece-e1f248c7a9b1.png)

tomayac:
The dump is just the saved original (save page including images). This definitely did work.

rviscomi:
Are the missing metrics the ones dependent on user input that are expected not to have any data in the lab?

rviscomi:
I think I found the cause of the discrepancy. You had initially written the queries against the `pages` dataset, which produced 0s for each unused feature. My [PR review feedback](https://github.com/HTTPArchive/bigquery/pull/84/files/c6ed6814d61674a8f51f1bcf1fad1c883cbdaade#r459662684) was to query the `blink_features` dataset instead, which limited the results to the specific features and produced no results for features having no data. You must have saved the JSON results to the CDN corresponding to your initial queries rather than the version of the queries after code review.

[...]

PR #224, opened 2020-09-24
Remove Fugu reports that require a user gesture

tomayac:
Also in response to #221.

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -1027,11 +1027,11 @@
 				]
 			}
 		},
-		"nativeFileSystem": {
-			"name": "(🧪 Origin trial) Native File System",
+		"fileSystemAccess": {
+			"name": "(🧪 Origin trial) File System Access",
 			"type": "%",
 			"downIsBad": true,
-			"description": "This metric tracks the percentage of pages that access files via the [Native File System API](https://web.dev/native-file-system/).",
+			"description": "This metric tracks the percentage of pages that access files via the [File System Access API](https://web.dev/file-system-access/).",
 			"histogram": {
 				"enabled": false
 			},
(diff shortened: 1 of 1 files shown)


tomayac:
Related PR: https://github.com/HTTPArchive/bigquery/pull/88.

config/reports.json
```diff
@@ -1317,38 +1317,22 @@
...
-			"webNFCRead",
-			"webNFCWrite",
-			"webSerial",
-			"nativeFileSystem",
 			"contentIndex",
 			"quicTransport",
 			"idleDetection",
-			"webSocketStream",
-			"shapeDetectionFace",
-			"shapeDetectionText"
```
rviscomi, line 1351: If these features can never be detected in HA, should we remove their metric definitions above and corresponding SQL files in https://github.com/HTTPArchive/bigquery/pull/88?

tomayac: If the collection methodology changes, could this table contain data in the future?

rviscomi: Theoretically, but I don't foresee HA simulating user gestures anytime soon.

tomayac: Let me axe them then. They're easy to re-establish should we need them in the future.

rviscomi (approved):

rviscomi:
@tomayac I've staged these changes to https://httparchive-staging.ue.r.appspot.com/reports/project-fugu and I'm getting the following error: `'NoneType' object is not subscriptable`. Maybe one deleted metric wasn't removed from the report config?

tomayac:
Ooops, sorry. I killed one too many, fixed in https://github.com/HTTPArchive/httparchive.org/commit/dec36c82981292a1acfd54ebd5b2e1e645628eab.

rviscomi:
Thanks! Looks good in staging. Deploying to prod now.

christianliebel:
@rviscomi Hi Rick, on prod (https://httparchive.org/reports/project-fugu) I can only see the data until August 1, 2020. Shouldn't the data from the September crawl also be there? Thanks!

rviscomi:
Normally yes the September data should be available by now, thanks for pointing that out. I think the issue may be related to #225. Once that's resolved I'll restart the metrics pipeline (or wait for it to start again at the end of the month)

In the context of the PWA chapter of the Web Almanac, I'd recommend sourcing the data from BigQuery so we can follow a consistent data pipeline methodology for all metrics/chapters. Along those lines, you should also be looking at August 2020 data for parity with other chapters.

PR #226, opened 2020-10-14
Update gettingstarted_bigquery.md

tpiros:
Missing end ` in query.


rviscomi (approved):
Sorry I didn't see this sooner, thanks for the fix @tpiros!

PR #227, opened 2020-10-23
Update changelog.json

rviscomi:
Log the mobile CPU throttling change that took effect in August 2020

### DIFF
--- a/static/json/changelog.json
+++ b/static/json/changelog.json
@@ -82,6 +82,11 @@
     {
         "date": 1596254400000,
         "title": "Faster mobile CPU throttling",
-        "desc": "WebPageTest changed its mobile CPU throttling from cgroups to Chrome's built-in throttling"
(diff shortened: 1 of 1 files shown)


static/json/changelog.json
```diff
@@ -78,5 +78,10 @@
         "date": 1544850000000,
         "title": "Synced URLs with entire CrUX corpus",
         "desc": ""
+    },
+    {
+        "date": 1596254400000,
+        "title": "Faster mobile CPU throttling",
```
pmeenan, line 84: Maybe "Mobile CPU Throttling Changed (Faster performance)"

static/json/changelog.json
```diff
@@ -78,5 +78,10 @@
...
+        "desc": "WebPageTest changed its mobile CPU throttling from cgroups to Chrome's built-in throttling"
```
pmeenan, line 85: Maybe be a bit more descriptive: "WebPageTest changed its mobile CPU throttling from cgroups to Chrome's built-in throttling. The cgroups had more overhead than expected when running in VM's and switching to Chrome's built-in throttling eliminated the extra overhead and resulted in more accurate throttling (which led to slightly faster performance)."

rviscomi: Thanks! Also added a link to the issue for more context.

PR #253, opened 2021-04-25
Add Lighthouse automatic checks

bazzadp:


bazzadp:
And after releasing:

Home page - 91/100/93/100

About page - 85/100/93/100

Reports - 98/100/93/100

Accessibility Report - 53/100/93/100

State of the Web Report - 47/100/93/100

Issue #250, opened 2021-04-24
Take learnings from Web Almanac site and apply to main httparchive site


bazzadp:
We learned a lot from making the Web Almanac site and it follows a lot of best practices, particularly in terms of Accessibility, Performance, SEO and automation to ensure we stay at the top of that.

We should bring those learnings here and align the web sites, which will also make them easier to maintain together.

Making a list so I don't forget anything:
- [x] - Move Virtual Env to `.venv` to prevent linking problems - #241 
- [x] - Add robots.txt and sitemap.xml - #243 
- [x] - Add GitHub SuperLinter - #244 
- [x] - Rename `master` branch to `main` - #245 
- [x] - Add CodeQL scanning for JS and Python files to supplement Super Linter - #246 and #247 
- [x] - Add test script and run it in an action #248
- [x] - Update all dev dependencies and automate #249  
- [ ] - Look at whether to remove Babel? Is it really needed as quite chunky an d adds a lot of dependencies
- [ ] - Look at whether to remove Bootstrap - do we really need it?
- [ ] - Might need another normalize CSS script if we do remove Bootstrap.
- [ ] - Add metadata for SEO reason (see also #34 )
- [ ] - Set `rel="canonical"` on reports to avoid query param versions getting indexed.
- [ ] - Set `noindex` when not using httparchive.org domain to avoid GCP version getting indexed
- [ ] - Add Lighthouse tests on pull request
- [ ] - Lighthouse review
- [ ] - Look at CSP
- [ ] - Look at colour contrast
- [ ] - Move to `em`/`rem` based sizes to allow rescaling.
- [ ] - Accessibility review
- [ ] - Performance review
- [ ] - SEO review
- [ ] - Add web-vitals script?
- [ ] - Refactor Python scripts to organise into a `server` directory
- [ ] - Add pytest (ideally with 100% coverage!)

Feel free to add to this as you see fit, and grab some if they seem of interest. I'll start will work my way through them anyway.

bazzadp:
Made quite a bit of progress with this, however the Accessibility improvements in particular have necessitated some changes:

- Some of the text had bad colour contrast, including the sub-headings:

![image](https://user-images.githubusercontent.com/10931297/116013841-24baef00-a62a-11eb-8856-53347037aa44.png)

and the table headings:

![image](https://user-images.githubusercontent.com/10931297/116013852-37cdbf00-a62a-11eb-934f-93ca89d252a2.png)

And our links:

[...]

PR #256, opened 2021-04-25
Resolve a number of accessibility issues

bazzadp:
Part of #250 

Resolves a number of accessibility issues raised by Axe, WAVE and Firefox accessibility tab


config/reports.json
```diff
@@ -1038,7 +1038,7 @@
     }
   },
   "chrome-ux-report": {
-    "name": "Chrome User Experience Report",
+    "name": "Chrome User Experience",
```
rviscomi, line 1041: "Report" is part of the product name

rviscomi (changes requested):

config/reports.json
bazzadp: The reason I removed it is cause some of the a11y tools were (not unreasonably) complaining about the confusion of two "Chrome User Experience Report" links, linking to two difference places:

image

The heading links to the HTTP Archive CrUX Report (The Chrome User Experience Report Report if you will!), while the second link goes to the CrUX home page.

It's a minor thing so can just revert and ignore it, unless you've any better suggestions on how to handle this?

bazzadp: There's also the fact I renamed the buttons so we wouldn't have all the buttons called "View the Report", but a more descriptive button:

image

If we revert "Chrome User Experience" back to "Chrome User Experience Report", then it's button will become "View the Chrome User Experience Report Report" which, while technically accurate is a bit of a mouthful!

Plus I didn't like the fact the original title wrapped when none of the others did 😁

rviscomi: How about renaming the report to `CrUX`? Shorter and allows "Report" after.

bazzadp: Done!

PR #271, opened 2021-04-30
Fix report dates

bazzadp:
The dates have stopped working on the site. Looks like the API has changed and `_get_next_page_response()` doesn't work as it used to and only returns the first page of results (2010 - 2012).

This slight tweak [as suggested in the official repo](https://github.com/googleapis/google-cloud-python/issues/920) works.

However it does mean launching the new site with all the tweaks and not sure if you've had a chance to review that yet @rviscomi ?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,11 @@
 version: 2
 updates:
   - package-ecosystem: "npm"
-    directory: "/src/"
+    directory: "/"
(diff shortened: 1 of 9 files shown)


bazzadp:
Test site here: https://20210430t144323-dot-httparchive.uk.r.appspot.com/reports/

rviscomi (approved):

Issue #283, opened 2021-05-06
April reports not available


rviscomi:
The reports on httparchive.org should have updated with April data by now. Needs investigation. @paulcalvano @bazzadp would either of you be interested to chase this down?

bazzadp:
Looks like the data is not there on Google Cloud Storage:

![image](https://user-images.githubusercontent.com/10931297/117266218-ece35100-ae4c-11eb-884f-a3346f640d85.png)

What puts it there?

A second issue is that lots of the reports are stuck at Dec 2020. Looking at that folder on Google Cloud Storage, there's lots of reports:

![image](https://user-images.githubusercontent.com/10931297/117266406-24ea9400-ae4d-11eb-9cd0-b694b7ba1f08.png)

However the March folder only have one report:

![image](https://user-images.githubusercontent.com/10931297/117266481-36cc3700-ae4d-11eb-9d21-d1791c1dae17.png)

February looks OK though.

Then there is a third issues! In that some of the reports are stuck at Dec 2020. Again this looks like the data being incorrect on Google Cloud Storage.

For example the [sampleSize report](https://httparchive.org/reports/state-of-the-web?start=2017_01_01&end=2021_02_01&view=list#numUrls) is based on data at this URL: https://storage.googleapis.com/httparchive/reports/numUrls.json
As you can see that's not updated since Dec 2020.

bazzadp:
Reports have been rerun and a monitoring script added. Will investigate more if May reports fail again.

rviscomi:
Thank you for chasing that down and adding tests to catch this sooner!

PR #285, opened 2021-05-07
Add test script for report data

bazzadp:
Adds a script and a GitHub action to test the reports exist for this month on the 28th of the month.

Opens an issue if they are not.

Hopefully avoids issues like #283 where we don't notice for a month or two.

Quite like having this completely outside of Google Cloud Platform/Storage as an independent test.


bazzadp:
Note I picked the first and last URL in each directory. And randomly picked Drupal as the lens.

Happy to test some more URLs or lens at same time. Easy enough to add as you can see.

PR #316, opened 2021-05-19
Fix annotations

bazzadp:
Fixes a couple of bugs noticed recently:

First up - the annotations are not showing up properly:

Annotations on timeseries charts

Looks like a bug introduced in v6.1.4 and I recently upgraded from v6.0.2 all the way to latest (9.1.0).
I can't actually see what's wrong here, but it's definitely to do with the clip path it adds. Overriding it solves the issues. Not the happiest with this fix but spent a long time at it and can't get a better one.

Secondly I noticed we were loading the exporting.js from CDN meaning we were loading the v6.0.2. version still. It appears to work despite the massive version gap but still - should be in sync. This moves it inline with the rest and also means the exports get the chart name rather than a generic chart.png.

### DIFF
--- a/config/last_updated.json
+++ b/config/last_updated.json
@@ -11,8 +11,8 @@
   },
   "/static/css/report.css": {
     "date_published": "2018-05-08T00:00:00.000Z",
-    "date_modified": "2021-05-19T00:00:00.000Z",
-    "hash": "46c7c17f5035a5209af3ddc5a2f1d4ea"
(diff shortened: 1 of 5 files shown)


bazzadp:
Found the documented, non-hacky way of doing this: https://api.highcharts.com/highstock/series.flags#:~:text=Defaults%20to%20undefined.-,clip,-%3A%20boolean

Much happier with this now so will merge and release.

PR #348, opened 2021-06-12
Update gettingstarted_bigquery.md

myakura:
the BigQuery pricing page has an ID for each major section including free tier.
this updates the link in the doc to that section so that people don't have to scroll or do find-in-page.


tunetheweb (approved):

tunetheweb:
Thanks @myakura !

PR #358, opened 2021-06-24
Add HTTP/3 report

tunetheweb:
Adds the new HTTP/3 report as created in https://github.com/HTTPArchive/bigquery/pull/111

HTTP/3 report

I've reached out to some HTTP/3 experts to see if they can explain that dip.

Also added some more explanation to HTTP/2 chart:

HTTP/3 chart

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -289,7 +289,21 @@
     "h2": {
       "name": "HTTP/2 Requests",
       "type": "%",
-      "description": "The percent of all requests in the crawl using HTTP/2.",
+      "description": "The percent of all requests in the crawl using HTTP/2. Note servers supporting HTTP/2 and HTTP/3, may use HTTP/2 initially due to the way the HTTPArchive starts with a fresh Chrome instance each time, but may use HTTP/3 on subsequent requests.",
+      "downIsBad": true,
+      "histogram": {
+        "enabled": false
+      },
+      "timeseries": {
+        "fields": [
+          "percent"
+        ]
+      }
+    },
+    "h3": {
+      "name": "HTTP/3 Support",
+      "type": "%",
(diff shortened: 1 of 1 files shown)


tunetheweb:
Also ran the timeseries and the 3 lens's manually and all worked fine. Wordpress looks interesting:

Wordpress HTTP/3 support

config/reports.json
```diff
@@ -289,7 +289,21 @@
...
+      "timeseries": {
+        "fields": [
+          "percent"
+        ]
+      }
+    },
+    "h3": {
+      "name": "HTTP/3 Support",
+      "type": "%",
+      "description": "The percent of all requests in the crawl which support HTTP/3. Note that, due to the way the HTTPArchive starts with a fresh Chrome instance each time, HTTP/3 will often not be used for initial connections until the browser is aware the server supports HTTP/3, which is why we measure support, rather than actual usage like we do for HTTP/2.",
```
rviscomi, line 306:
```suggestion
      "description": "The percent of all requests in the crawl which support HTTP/3. Note that, due to the way the HTTP Archive starts with a fresh Chrome instance each time, HTTP/3 will often not be used for initial connections until the browser is aware the server supports HTTP/3, which is why we measure support, rather than actual usage like we do for HTTP/2.",
```

config/reports.json
```diff
@@ -289,7 +289,21 @@
     "h2": {
       "name": "HTTP/2 Requests",
       "type": "%",
-      "description": "The percent of all requests in the crawl using HTTP/2.",
+      "description": "The percent of all requests in the crawl using HTTP/2. Note servers supporting HTTP/2 and HTTP/3, may use HTTP/2 initially due to the way the HTTPArchive starts with a fresh Chrome instance each time, but may use HTTP/3 on subsequent requests.",
```
rviscomi, line 292:
```suggestion
      "description": "The percent of all requests in the crawl using HTTP/2. Note that servers supporting HTTP/2 and HTTP/3 may use HTTP/2 initially due to the way the HTTP Archive starts with a fresh Chrome instance each time, but may use HTTP/3 on subsequent requests.",
```

rviscomi (approved):

PR #396, opened 2021-07-29
Add rank lenses

tunetheweb:
Once https://github.com/HTTPArchive/bigquery/pull/121 lands and we run the last of the lenses we should add to the report. This PR does that:

Lens drop down

Question is whether to drop the old lenses or not? I vote yes to drop from GUI (i.e. update this PR to remove them), but leave them running in background for another month in case of push back. Then drop from bigquery repo if no complaints. Thoughts?

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -23,6 +23,18 @@
     },
     "wordpress": {
       "name": "WordPress"
+    },
+    "top1k": {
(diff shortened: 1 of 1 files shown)


tunetheweb:
As discussed on slack we'll keep the other lenses for now.

Issue #391, opened 2021-07-26
Add a link to the CWV Tech Report


rviscomi:
Make it easier to discover the [Core Web Vitals Technology Report](https://datastudio.google.com/reporting/55bc8fad-44c2-4280-aa0b-5f3f0cd3d2be/page/RNEOC) by linking to it from the HA website.

Not sure of the best place for it. Maybe as a special item in the "Reports" navbar list and landing page?

PR #441, opened 2021-09-18
Add CWV Tech report link

tunetheweb:
Fixes: #391 

To differentiate it slightly, I've given it a white background (and also removed this white background from the PWA report as didn't see why it needed differentiating to be honest):

image

It also appears in the menu, but not differentiated:

Could differentiate it some more, but couldn't think of anything better for now and, while we only have one external report, don't think it's worth giving it a whole new section like we did for Web Almanac.


main.py
```diff
@@ -161,7 +161,8 @@ def reports():
 @app.route('/reports/<report_id>')
 def report(report_id):
     report = report_util.get_report(report_id)
-    if not report:
+    external_reports = report_util.get_external_reports()
+    if not report or report_id in external_reports:
```
rviscomi, line 165: Could we 302 redirect to the external report URL instead? That would also provide a more memorable URL for the dashboard.

rviscomi (changes requested):

main.py
tunetheweb: Good idea.

Why 302 (temporary) over 301 (permanent) though? Or you think it might eventually be hosted on httparchive.org?

Implemented 301 for now in https://github.com/HTTPArchive/httparchive.org/pull/441/commits/1c3e6c5678774099827753dfe9eaed0dea05f7f9 but easy to update if you want.

rviscomi: We want to be able to change the URL if needed. My understanding of 301 was that it could cause the old URL to get stuck in some caches.

tunetheweb: Fair enough. Not too concerned about difference to be honest as don't want this page indexed from here, so as you say, safer to go 302. Changed!

Happy to release now?

rviscomi: Yes, LGTM. Thanks!

rviscomi (approved):

Issue #527, opened 2022-01-17
pip & pip3


thehassantahir:
![Screenshot from 2022-01-17 17-23-39](https://user-images.githubusercontent.com/54933420/149768946-c69e8aa6-bad2-407a-80bd-ada8c2c28232.png)

If you are suggesting to use _Python3.8_ to run the program then you should use `pip3` instead of `pip` to install the environments, It will help the community to start the env without any errors (which defiantly avoid the issue reporting time).

Let me know if i am lacking anywhere 🙂

tunetheweb:
For Python 3 users does `pip` not map to `pip3` anyway? And, given we suggest developing the site in a Python 3 virtual env, I'm not sure why being explicit with `pip3` is needed?

But not a Python expert so happy to hear reasons here, as think it would equally work with `pip3` too.

thehassantahir:
@tunetheweb As far as i understand for Python 3 usage `pip` does not map to `pip3`, It will show an error like:
`pip not recognized` if you try to use `pip` with python 3 programs, I have tried it so many times for python 3 programs and all time the `pip3` satisfy the process.
Best practice is to use the supporting technology to avoid errors & it all depends on the developers, I am in a learning phase too! 😄

tunetheweb:
Are you using a virtualenv as suggested in the screenshot in your first comment?

If so you will see they are the same:

```
% source .venv/bin/activate
% pip --version
pip 21.3.1 from /Users/barry/sources/httparchive.org/.venv/lib/python3.8/site-packages/pip (python 3.8)
% pip3 --version
pip 21.3.1 from /Users/barry/sources/httparchive.org/.venv/lib/python3.8/site-packages/pip (python 3.8)
```

If you are not using a virtualenv then you are not following the instructions and will experience other issues too.

thehassantahir:
I am using virtual env but somehow my `pip` is not recognizing Python 3 

Reason:
I have two python versions installed and that's what I was talking about if someone is using to different python versions pip will not automatically map to pip3 because they both have their dedicated directory according to versions.
```
pip --version
pip 20.3.4 from /usr/lib/python/dist-packages/pip (python 2.7)  
pip3 --version  
pip 21.3.1 from /usr/lib/python3/dist-packages/pip3 (python 3.9)
```
But inside virtual env I can see your `pip` is just good & I got the point how to merge them.
@tunetheweb According to you, You are saying perfectly fine (No need to change `pip` to `pip3`)

[...]

PR #528, opened 2022-01-17
replaced pip with pip3

thehassantahir:
Issue highlighted #527


tunetheweb:
As discussed in #527 this should not be necessary when using VirtualEnvs.

PR #615, opened 2022-06-22
Add changelog about May 2022 response bodies not being available

### LINKED ISSUE #614: Add a change log entry about response bodies in May 2022
Due to a crawl config error, we don't have response bodies in the May 2022 dataset. Add this to the change log.

tunetheweb:
Fixes #614 

Note this also adds an optional `displayInTimeSeries` boolean flag that when set to `false` is excluded from the timeseries charts.

### DIFF
--- a/src/js/timeseries.js
+++ b/src/js/timeseries.js
@@ -259,6 +259,8 @@ const getFlagSeries = () => loadChangelog().then(data => {
       desc: change.desc
     };
   });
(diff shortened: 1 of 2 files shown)


tunetheweb:
Timeseries still only goes up to Q:

image

Changelog has them all:

image

rviscomi (approved):

Issue #629, opened 2022-07-27
Improve social metadata


rviscomi:
Spotted a few grammar mistakes in the social description

<img width="525" alt="image" src="https://user-images.githubusercontent.com/1120896/181364075-c050bf17-1f55-410c-9935-5cf170183471.png">

PR #630, opened 2022-07-27
Update description

rviscomi:
Fixes https://github.com/HTTPArchive/httparchive.org/issues/629

### DIFF
--- a/templates/about.html
+++ b/templates/about.html
@@ -3,7 +3,7 @@
 
 {% block title %}About HTTP Archive{% endblock %}
 
-{% block description %}The HTTP Archive Tracks how the web is built by periodically crawl the top sites on the web and record detailed information about fetched resources, used web platform APIs and features, and execution traces of each page.{% endblock %}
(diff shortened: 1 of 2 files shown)


tunetheweb:
Why dropping of "The"? It's what's on our [home page](https://httparchive.org/):

image

Reads weird to me without it.

rviscomi:
Reverted "the"

templates/about.html
```diff
@@ -3,7 +3,7 @@
 
 {% block title %}About HTTP Archive{% endblock %}
```
tunetheweb, line 4:
```suggestion
{% block title %}About the HTTP Archive{% endblock %}
```

Should we review all the titles?

templates/index.html
```diff
@@ -3,7 +3,7 @@
 
 {% block title %}HTTP Archive{% endblock %}
```
tunetheweb, line 4:
```suggestion
{% block title %}The HTTP Archive{% endblock %}
```

tunetheweb (changes requested):

tunetheweb (approved):

PR #640, opened 2022-08-17
Redirect legacy to bare domain

rviscomi:
DNS for the legacy subdomain is set up to direct to GAE. This adds support for redirecting it to the bare domain, similar to www and beta subdomains.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -62,7 +62,11 @@ legacy_util = Legacy(faq_util)
 def redirect_www():
     """Redirect subdomain requests to bare domain."""
     urlparts = urlparse(request.url)
-    if urlparts.netloc == 'www.httparchive.org' or urlparts.netloc == 'beta.httparchive.org' or urlparts.netloc == 'legacy.httparchive.org':
+    if (
+        urlparts.netloc == "www.httparchive.org"
+        or urlparts.netloc == "beta.httparchive.org"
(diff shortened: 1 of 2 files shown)


server/__init__.py
```diff
@@ -60,9 +60,9 @@ def get_send_file_max_age(self, name):
 
 @app.before_request
 def redirect_www():
-    """Redirect www and beta requests to bare domain."""
+    """Redirect subdomain requests to bare domain."""
     urlparts = urlparse(request.url)
-    if urlparts.netloc == 'www.httparchive.org' or urlparts.netloc == 'beta.httparchive.org':
+    if urlparts.netloc == 'www.httparchive.org' or urlparts.netloc == 'beta.httparchive.org' or urlparts.netloc == 'legacy.httparchive.org':
```
tunetheweb, line 65:
```suggestion
    if (
        urlparts.netloc == "www.httparchive.org"
        or urlparts.netloc == "beta.httparchive.org"
        or urlparts.netloc == "legacy.httparchive.org"
    ):
```

Let's also add a test for this.

tunetheweb (changes requested):

tunetheweb (approved):

tunetheweb:
Deployed that change so good to switch over in DNS whenever you want to.

PR #673, opened 2022-11-30
Update project name

aaemnnosttv:
The current instructions don't work when using the given project name and results in a rather cryptic error.

image
image

As shown in https://discuss.httparchive.org/t/bigquery-broken-link/2366/11 (and confirmed in my own set up), the project name must be all lower case.

This PR corrects the instruction as well as emphasizes the importance of the case for the HTTP Archive project data to be added successfully.


tunetheweb (approved):
Ta!

PR #728, opened 2023-05-14
[WIP] Add core web vitals report type

sarahfossheim:
## Note:
- **! Work in progress / Prototype**
- **The PR can be reviewed already, but the outcome is not a production-ready app :)** 

**The PR is focused on:**
- Configuring the new CWV tech report
- Putting in place 
- Adding configuration and template and component structure for the tech report
- Load in the data
- Creating a basic frame and way of displaying the data as a proof of concept

## Included:
- Text, labels, and other config in the `techreport.json` config file
- Landing, Drilldown, Comparison page template files
- Filters (apps, region, rank) can be set through selects, need to press update button to update the filter 
    - This also updates the URL
    - The correct **page skeleton is drawn on the server-side** based on the filters (URL) and the configuration (techreport.json)
    - The geo and rank filters carry over between the pages
    - The APP name is case sensitive and needs to be formatted the same as in the API
    - The category dropdown filters the technology dropdown
    - On the comparison page it’s possible to select multiple technologies
- The **data is fetched and drawn client-side** 
    - On the client we fetch the data through .json files in one global class
    - When the data is fetched, it gets processed and then the HTML gets updated with it
- The data can be **further filtered without requiring a new API call** by the following criteria:
    - Client, on the page level
    - Sub-metric, on the component level
- The following **components** are added:
    - Report navigation, to switch between the different pages
    - Technology (+category), rank, geo filters
    - Summary cards on the drilldown page
    - Summary table on the comparison page
    - Timeseries per metric (Good CWVs, lighthouse scores, page weight, adoption)
        - Consists of line chart (Highcharts), summary stats, and table alternative
        - Possible to select sub-metric that updates all the contents of the timeseries
    - CWV sub-metric comparison table
- Start of the styling the pages and components
- More info in: `docs/cwv_tech_report.md`

## To be added
PR can be reviewed even though these are not added yet:
- [ ] Charts in the comparison page The comparison page currently consists only of tables, add the Highcharts timeseries there too
- [ ] Code cleanup
    - [ ] Make summary cards and table follow the same format as the charts
    - [ ] Fetch all text labels from a settings file (easier to update in one central location)
    - [ ] Continue documentation
    - [ ] Linting
- [ ] Bugfixes
    - [ ] Category filter doesn’t work when adding new techs in the comparison view

### In follow-up PR(s) this autumn:
- Switch from JSON files to the new API
- Styling of the filters
- Improved styling of the page and charts
- UX improvements & improved search 
- Copywriting / correct text labels
- Accessibility & performance updates
- Additional charts based on the metrics we have
- …
- See Notion / Figma

### Later:
What we know of already:
- Improved functionality and styling
- Additional metrics and charts:
    - Median CWVs
    - Distribution of the CWV values
    - Breakdown / compare / explore by rank
    - Lighthouse audits / opportunities
    - … 
- …
- See Notion / Figma


config/techreport.json
```diff
@@ -0,0 +1,78 @@
...
+  "summary": "The Core Web Vitals Technology Report is a dashboard combining the powers of real-user experiences in the [Chrome User Experience Report (CrUX)](https://developers.google.com/web/tools/chrome-user-experience-report/) dataset with web technology detections available in HTTP Archive, to allow analysis of the way websites are both built and experienced.",
+  "metrics": [],
+  "pages": {
+    "landing": {
+      "id": "landing",
+      "title": "Technology Report",
+      "subtitle": "Report",
+      "description": "This is placeholder text about how the report works",
+      "sections": [
+        {
```
sarahfossheim, line 13: @tunetheweb something to discuss: to which extend do we want to configure the pages / sections / components in a json file vs. building the page structure directly in the techreport templates?

config/mock_responses/adoption.json
```diff
@@ -0,0 +1,39 @@
+{
+  "results": [
+    {
+      "technology_name": "Wordpress",
+      "origins": [
```
sarahfossheim, line 5: @tunetheweb is it correct that we'll have one api per metric? or will adoption, metadata, CWVs, lighthouse audits, etc. all come as a result from the same call?

config/mock_responses/good_vitals.json
```diff
@@ -0,0 +1,33 @@
+{
+  "results": [
+    {
+      "technology": "wordpress",
+      "dates": [
+        {
+          "date": "2022-04-22T00:00:00.000Z",
+          "tested": 10,
```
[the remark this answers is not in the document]
tunetheweb: Going to try to play with an alternative approach this week and next. Will let you know how it goes.

config/mock_responses/good_vitals.json
```diff
@@ -0,0 +1,33 @@
...
+          "good_overall": {
+            "desktop": 4,
+            "mobile": 7,
+            "dataset": 6
```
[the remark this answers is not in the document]
tunetheweb: Not started playing with it, but I think we should have one API call per technology. And include dates, mobile, desktop, Core Web Vitals, Lighthouse...etc. in it.

`all` will also count as a technology too.

config/techreport.json
tunetheweb: I'm OK with NOT putting too much into JSON. It works well for the other reports where they are all basically copies of each other with different config. But this will be it's own, individual one so quite different.

Then again, having the labels in JSON would allow us to potentially offer the report in alternative languages. But should be easy enough to convert the hardcoded labels to config later.

config/mock_responses/adoption.json
tunetheweb: As per above, I've not started playing with it, but I think we should have one API call per technology. And include dates, mobile, desktop, Core Web Vitals, Lighthouse...etc. in it.

`all` will also count as a technology too.

config/mock_responses/adoption.json
```diff
@@ -0,0 +1,35 @@
...
+          "date": "2022-04-22T00:00:00.000Z",
+          "origins": {
+            "mobile": 16345,
+            "desktop": 735
+          }
+        }
+      ]
+    }
+  ]
+}
```
tunetheweb, line 35: One minor nit. Do you have editorconfig set up on your IDE? We have a config file to auto add final line breaks (plus other things) to stop GitHub flagging these:

image

config/mock_responses/adoption.json
sarahfossheim: As I'm configuring some of the components now, it feels like it would indeed be easier to have one API call per tech and then include everything in there. 

Would also be useful if the results will be formatted more or less the same way, so we can do something reusable along the lines of `tech.results[metric][selected_client]` (with metric being eg. _origins_, _good_cwvs_, _lighthouse_performance_, etc) to render the values.

PR #752, opened 2023-07-06
LoAF origin trial

rviscomi:
According to Search Console, we have an INP problem on our report pages. I think this is a good opportunity to try out the new LoAF origin trial in Chrome ~115~116 to see if it helps us narrow the problem down. While it might not be _necessary_ for debugging, it's a good opportunity for us to dogfood the API.

I've registered all HA subdomains for the trial and this PR adds the meta token.

image

### DIFF
--- a/config/last_updated.json
+++ b/config/last_updated.json
@@ -31,8 +31,8 @@
   },
   "/static/js/exporting.js": {
     "date_published": "2021-05-19T00:00:00.000Z",
-    "date_modified": "2023-06-23T00:00:00.000Z",
-    "hash": "f36e2313c705b44e0779f7c34d9e34e0"
+    "date_modified": "2023-08-16T00:00:00.000Z",
+    "hash": "a2a9eaccff8895f1127a71d1821a18e4"
   },
   "/static/js/exporting.js.map": {
     "date_published": "2021-09-19T00:00:00.000Z",
@@ -46,23 +46,23 @@
   },
   "/static/js/highcharts-more.js": {
     "date_published": "2021-04-26T00:00:00.000Z",
-    "date_modified": "2023-06-23T00:00:00.000Z",
-    "hash": "78c6b81245d874b48c35f1c15376c507"
+    "date_modified": "2023-08-16T00:00:00.000Z",
+    "hash": "38d4924da68d0fdfe145f059d78840f1"
   },
   "/static/js/highcharts-more.js.map": {
     "date_published": "2021-09-19T00:00:00.000Z",
-    "date_modified": "2023-06-23T00:00:00.000Z",
-    "hash": "4d58940570fe351aebc821a4db5dac17"
+    "date_modified": "2023-08-16T00:00:00.000Z",
+    "hash": "8d31b26d9fd75b9f2749cf0fc652fb34"
   },
   "/static/js/highstock.js": {
     "date_published": "2021-04-26T00:00:00.000Z",
-    "date_modified": "2023-06-23T00:00:00.000Z",
-    "hash": "f470e10501661f6de5214aa01a7efff7"
+    "date_modified": "2023-08-16T00:00:00.000Z",
+    "hash": "a0f4226f7c89b57f57921439255b188b"
   },
   "/static/js/highstock.js.map": {
     "date_published": "2021-09-19T00:00:00.000Z",
-    "date_modified": "2023-06-23T00:00:00.000Z",
-    "hash": "515d855bc02b96ec1652a81694fb93b7"
+    "date_modified": "2023-08-16T00:00:00.000Z",
+    "hash": "6a6de9a402c29ea5b76e30b5eed6ddd9"
   },
   "/static/js/histogram.js": {
     "date_published": "2018-05-08T00:00:00.000Z",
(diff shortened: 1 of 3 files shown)


tunetheweb:
No issues adding but what’s the use case for enabling for everyone given we’re not sending LoAF data to analytics? Or is that planned for a future PR?

I guess at least it enables it without having to enable all the other Experimental Web Platform features?

rviscomi:
Yeah I'd like to also include it in our analytics. I'd need more time to get familiar with it but I can follow that up in a future PR.

tunetheweb:
I'm also supposed to be writing a blog post on LoAF. In which I should consider questions like this! :-)

rviscomi:
Looking forward to it! At least in our case, since we already have real-user INP issues, it'd be really interesting to see how that maps back to the real-user LoAF diagnostics. Setting the flag locally is probably good enough for most curious people.

rviscomi:
Not sure if it's too bleeding edge for web-vitals.js, but I filed a FR to add LoAF support to the attribution build: https://github.com/GoogleChrome/web-vitals/issues/363

rviscomi:
Working on this actually helped uncover a gap in the API: https://bugs.chromium.org/p/chromium/issues/detail?id=1473663 🙌

config/last_updated.json
```diff
@@ -31,8 +31,8 @@
   },
   "/static/js/exporting.js": {
     "date_published": "2021-05-19T00:00:00.000Z",
-    "date_modified": "2023-06-23T00:00:00.000Z",
-    "hash": "f36e2313c705b44e0779f7c34d9e34e0"
+    "date_modified": "2023-08-16T00:00:00.000Z",
+    "hash": "a2a9eaccff8895f1127a71d1821a18e4"
```
tunetheweb, line 35: FYI you don't need to update any of these (they are done with a GitHub Action just prior to release), and doing so is likely to lead to merge conflicts if other PRs are open at the same time and they also do this.

src/js/send-web-vitals.js
```diff
@@ -80,32 +128,32 @@ function sendWebVitals() {
...
-        navigation_type: navigationType,
-      }, overrides));
+    gtag('event', name, params);
+
+  }
 
+  // Monitor LoAFs
+  if (PerformanceObserver.supportedEntryTypes.includes('long-animation-frame')) {
+    new PerformanceObserver(entries => {
+      longAnimationFrames = longAnimationFrames.concat(entries.getEntries());
```
tunetheweb, line 152: We never clear down the `longAnimationFrames` value. That means this will just grow and grow and grow as long as the user remains on the page. At the same time, you cannot clear down until you know the INP entry (which won't be reported until later), so tricky. This is one of the complications I was talking about.

Would you be open to additionally monitoring all long LoAFs (>= 150ms)? Or all long LoAFs with interactions? That way we could compare the two approaches.

tunetheweb: An alternative approach is to not use a Performance Observer for the LoAFs and just look them up in the performance timeline:

```
performance.getEntriesByType('long-animation-frame');
```

Now, they aren't guaranteed to hang round forever (which is why PO is generally preferred), so may get some blank entries for long lived pages, but easier than managing this in our own JavaScript.

rviscomi: For now I was planning to monitor the `debug_loaf_length` value and take action if it grows beyond a reasonable amount for most users. If that works for you we can pick the threshold.

tunetheweb: So for now the [Performance Timeline buffer is set to 200 in Chrome](https://github.com/chromium/chromium/blob/54a5d0c487f5d388613f0b3fca30dcb05a0a5163/third_party/blink/renderer/core/timing/performance.cc#L256) so could go with that, which should surely be enough for most pages.

Then again if doing that, then why just use `performance.getEntriesByType('long-animation-frame');`?

rviscomi: Didn't see your latest comment before reply. Yes that seems like a much better solution.

tunetheweb: Added that to the LoAF post too.

rviscomi: PTAL

src/js/send-web-vitals.js
```diff
@@ -1,7 +1,48 @@
...
+    const loafAttribution = {
+      total: 0
+    };
+
+    const longAnimationFrames = performance.getEntriesByType('long-animation-frame');
+    longAnimationFrames.filter(loaf => {
+      // LoAFs that intersect with the event.
+      return entry.startTime < (loaf.startTime + loaf.duration) && loaf.startTime < (entry.startTime + entry.duration);
+    }).forEach(loaf => {
+      loaf.scripts.forEach(script => {
```
tunetheweb, line 23: I wonder if we should also capture the overall LoAF timings?: https://gist.github.com/noamr/2d4e2bdd48661fa39e32283efecaf418#file-whynp-js-L10-L17

And should we put the script stuff in it's own `longestScript` object within the `loafAttribution` object to more accurately reflect what it is showing?

It would also be good to generalise this into a `summariseLoAF` function so could also capture all LoAFs separately with a PO, as an alternative approach to compare whether targeting INP LoAF or LoAFs in general is most useful:

```js
  // Monitor LoAFs above a 150ms threshold
  if (PerformanceObserver.supportedEntryTypes.includes('long-animation-frame')) {
    new PerformanceObserver(entries => {
      for (const entry of entries.getEntries()) {
        if (entry.duration > 150) {
          gtag('event', 'LoAF', summariseLoAF(entry));
        }
      }
    }).observe({type: 'long-animation-frame', buffered: true});
  }
```

rviscomi: Done

rviscomi:
Deployed live and the LoAF data appears to be successfully flowing
image

PR #766, opened 2023-09-07
Remove font awesome JS

tunetheweb:
We don't make a huge use of Font Awesome (3 icons on report pages, and multiple icons on the reports page) and it's pretty static. But currently we're loading the huge 245 kB (680 kB uncompressed) JavaScript library for every page, just for these small use cases.

This appears to be affecting our INP.

This PR inlines the few icons we use. We also add the use of Font Awesome icons to our README as we are still using these icons.


rviscomi (approved):
LGTM. Could you add a line to the README instructing how to get the `path` for any new icons someone may want to add in the future?

tunetheweb:
> LGTM. Could you add a line to the README instructing how to get the `path` for any new icons someone may want to add in the future?

Done. I copied from the old paths, but they also provide the paths on their website. The `width` isn't as easy to get without installing FA and seeing what it used, but I set that just to avoid any visual differences so not as important.

PR #797, opened 2024-01-17
Wait to render charts until visible

### LINKED ISSUE #796: Fix slow Interaction to Next Paint on the httparchive.org website
According to CrUX data in [PSI](https://pagespeed.web.dev/analysis/https-httparchive-org-reports-state-of-the-web/jmp5k3k488?form_factor=mobile) and Search Console, httparchive.org has a problem with INP on mobile:

image

image

Unfortunately, we don't have URL-level CrUX data beyond the home page, which is relatively simple and already performs well on INP. However, we do collect our own RUM data through web-vitals.js, Google Analytics, and BigQuery, so we can analyze how each page template performs:

```sql
SELECT
  REGEXP_EXTRACT(page_path, r'^(/[^/]*)') AS subdir,
  device.category AS device_type,
  COUNT(0) AS freq,
  COUNTIF(metric_value <= 200) / COUNT(0) AS pct_good_inp,
  APPROX_QUANTILES(metric_value, 1000 IGNORE NULLS)[OFFSET(750)] AS p75_inp
FROM
  `httparchive.analytics_386559427.web_vitals`
WHERE
  metric_name = 'INP' AND
  metric_value IS NOT NULL AND
  httparchive.analytics_386559427.FILTER_PREVIOUS_N_DAYS(event_date, 28, FALSE) AND
  device.category IN ('desktop', 'mobile') AND
  REGEXP_EXTRACT(page_path, r'^(/[^/]*)') IN ('/', '/reports', '/faq', '/about')
GROUP BY
  subdir,
  device_type
ORDER BY
  freq DESC
```

subdir | device_type | freq | pct_good_inp | p75_inp
-- | -- | -- | -- | --
/reports | desktop | 1962 | 92% | 56
/ | desktop | 808 | 97% | 24
/reports | mobile | 402 | 65% | 344
/ | mobile | 227 | 94% | 104
/faq | desktop | 86 | 93% | 24
/faq | mobile | 41 | 98% | 96
/about | desktop | 36 | 94% | 24
/about | mobile | 18 | 100% | 80

The results confirm that the `/reports` pages are dragging down the origin-level INP performance on mobile.

And in the few cases that we're able to get [LoAF attribution](https://github.com/HTTPArchive/httparchive.org/pull/752) out of the interaction, the RUM data strongly suggests that the issue has to do with chart initialization:

```sql
SELECT
  param.value.string_value AS loaf_source,
  COUNT(0) AS freq,
  APPROX_QUANTILES(metric_value, 100 IGNORE NULLS)[OFFSET(50)] AS median_inp
FROM
  `httparchive.analytics_386559427.web_vitals`,
  UNNEST(event_params) AS param
WHERE
  metric_name = 'INP' AND
  device.category = 'mobile' AND
  metric_value > 200 AND
  param.key = 'debug_loaf_script_source' AND
  httparchive.analytics_386559427.FILTER_PREVIOUS_N_DAYS(event_date, 28, FALSE)
GROUP BY
  loaf_source
ORDER BY
  freq DESC
```

loaf_source | freq | median_inp
-- | -- | --
`https://httparchive.org/static/js/timeseries.js?v=1499b116cf44f1152d064469eaeade73:0` | 10 | 880
`https://httparchive.org/static/js/report.js?v=63d4b296d1aba99875379b0d4d853128:2128` | 8 | 288
`onclick@https://httparchive.org/static/js/exporting.js?v=44a755704dad3fecd7b9d1022b9c6cb5:11759` | 5 | 264
`https://httparchive.org/static/js/web-vitals.js?v=0fa371ef0624041ff02cf6458e74441a:2437` | 3 | 240
`onContainerTouchStart@https://httparchive.org/static/js/highstock.js?v=c6744f333bf6edfeaed1200bd93a5` | 3 | 432
`r@https://httparchive.org/static/js/highstock.js?v=c6744f333bf6edfeaed1200bd93a537d:326741` | 2 | 1600
`https://httparchive.org/static/js/web-vitals.js?v=0fa371ef0624041ff02cf6458e74441a:1761` | 2 | 216
`onContainerTouchMove@https://httparchive.org/static/js/highstock.js?v=c6744f333bf6edfeaed1200bd93a53` | 2 | 216
`h@https://httparchive.org/static/js/highstock.js?v=c6744f333bf6edfeaed1200bd93a537d:21896` | 1 | 6856
`createHTML@https://httparchive.org/static/js/highstock.js?v=c6744f333bf6edfeaed1200bd93a537d:25273` | 1 | 1520
`Ux@https://www.googletagmanager.com/gtag/js?id=G-D66RRD8GF2:186103`| 1 | 1032
`r@https://httparchive.org/static/js/highstock.js?v=c6744f333bf6edfeaed1200bd93a537d:251811` | 1 | 1592
`t@https://httparchive.org/static/js/web-vitals.js?v=0fa371ef0624041ff02cf6458e74441a:2254` | 1 | 384
`https://httparchive.org/reports/state-of-the-web:0` | 1 | 832

Also supported by the interaction event timing data:

```sql
SELECT
  CEIL(param.value.int_value / 5000) * 5 AS interaction_time,
  COUNT(0) AS freq
FROM
  `httparchive.analytics_386559427.web_vitals`,
  UNNEST(event_params) AS param
WHERE
  metric_name = 'INP' AND
  device.category = 'mobile' AND
  STARTS_WITH(page_path, '/reports') AND
  metric_value > 200 AND
  param.key = 'debug_time' AND
  httparchive.analytics_386559427.FILTER_PREVIOUS_N_DAYS(event_date, 28, FALSE)
GROUP BY
  interaction_time
ORDER BY
  interaction_time
```

image

The histogram of interaction times shows 49% of slow INP interactions on mobile occurring within the first 10 seconds.

We can also see what users are interacting with when the slow INP occurs:

```sql
SELECT
  param.value.string_value AS target,
  COUNT(0) AS freq,
  APPROX_QUANTILES(metric_value, 100 IGNORE NULLS)[OFFSET(50)] AS median_inp
FROM
  `httparchive.analytics_386559427.web_vitals`,
  UNNEST(event_params) AS param
WHERE
  metric_name = 'INP' AND
  device.category = 'mobile' AND
  STARTS_WITH(page_path, '/reports') AND
  metric_value > 200 AND
  param.key = 'debug_target' AND
  httparchive.analytics_386559427.FILTER_PREVIOUS_N_DAYS(event_date, 28, FALSE)
GROUP BY
  target
HAVING
  freq > 1
ORDER BY
  freq DESC
```

target | freq | median_inp
-- | -- | --
`(not set)` | 28 | 672
`#toc>li>a` | 19 | 728
`html` | 7 | 1032
`#toc>li` | 4 | 1920
`#update` | 4 | 424
`div.container>div.col-lg-10.col-md-9.col-sm-9.col-xs-2.clearfix>button.visible-xs.hamburger>svg` | 4 | 656
`#toc>li>a>font>font` | 3 | 304
`#fontDisplay>button.btn.show-hide` | 3 | 288
`body>main>section.container>div.report.row>div.container>div.col-lg-7.col-sm-6.col-md-6.col-xs-12>p` | 3 | 440
`#numUrls>button.btn.show-hide` | 3 | 352
`#report-header>div.container` | 2 | 688
`div.container>div.col-lg-10.col-md-9.col-sm-9.col-xs-2.clearfix>button.visible-xs.hamburger>svg>path` | 2 | 1608
`#bytesJs>button.btn.show-hide` | 2 | 240
`#grid-view` | 2 | 208
`#permalink` | 2 | 272
`#report-metrics` | 2 | 1520

The target is null 31% of the time, but the list of chart links (`#toc`) shows up 25% of the time.

So knowing which pages have slow interactions, when, and how, I'm able to consistently reproduce slow INP like this: (https://trace.cafe/t/MAB4nLPUua with 4x CPU throttling)

image

Looking more closely at the 1.7s long task, it seems like all 9 of the timeseries charts are rendering in a single blocking operation at startup.

In my first attempt to fix this, I tried breaking it up into 9 separate tasks with `scheduler.yield` (falling back to `setTimeout`): (https://trace.cafe/t/O3d6sBJz6N)

image

It didn't quite solve the problem, because each rendering operation is still its own long task and we have no control over the performance of the Highcharts library. The interaction above is only barely under the 200ms threshold, so this didn't seem like a great solution.

It's also possible to incur a very slow INP even though the tasks are ~200ms each: (https://trace.cafe/t/MWIkIj0ZGY)

image

I managed this 1.1s input delay by interacting with the Start/End Date `select` lists, which apparently have some blocking behavior caused by the system UI.

So the solution I came up with is to completely defer rendering the charts until they're in view. That unblocks the main thread during the time when users are most likely to interact with the page. And when they do click the TOC to jump to a chart, they will have already interacted with the page before the long task to render the chart begins, avoiding the input delay. It's still possible to interact with the page again while it's loading (eg scroll it into view then interact) but it's much less likely.

image

[This trace](https://trace.cafe/t/r2PZfm9pev) shows a 16ms INP when clicking the TOC link, followed immediately by the long task to render the chart that was just scrolled into view.

This approach does have the downside of the charts not quite being ready to go when they're scrolled into view, but I think it's a good tradeoff for the INP improvements. Also, now the user gets to see the animation when the chart is initialized, so maybe it's a win-win.

I'll submit a PR to defer the charts and monitor the RUM data after the release to verify that the issue was resolved.

rviscomi:
Progress on #796 

Uses IntersectionObservers to defer rendering histogram and timeseries charts until they're in the viewport

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -58,6 +58,28 @@ export const drawMetricSummary = (options, client, value, isMedian=true, change=
   }
 };
 
+/**
+ * Invokes the callback when the element is visible for the first time.
+ * 
+ * Rendering all charts immediately at startup causes long tasks and slow INP.
+ * Instead, wait until the chart is in view before rendering.
+ * Rendering a single chart might still create a long task,
+ * but it'll be less likely to contribute to INP.
+ * 
+ * @param {HTMLElement} element 
+ * @param {Function} callback 
+ */
+export function callOnceWhenVisible(element, callback) {
+  new IntersectionObserver((entries, observer) => {
+    if (!entries[0].isIntersecting) {
+      return;
+    }
+
+    observer.unobserve(element);
+    callback();
+  }).observe(element);
+}
+
 const getQueryUrl = (metric, type) => {
   const URL_BASE = 'https://github.com/HTTPArchive/bigquery/blob/master/sql';
   if (type === 'timeseries') {
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -60,14 +60,14 @@ export const drawMetricSummary = (options, client, value, isMedian=true, change=
 
 /**
  * Invokes the callback when the element is visible for the first time.
- * 
+ *
  * Rendering all charts immediately at startup causes long tasks and slow INP.
  * Instead, wait until the chart is in view before rendering.
  * Rendering a single chart might still create a long task,
  * but it'll be less likely to contribute to INP.
- * 
- * @param {HTMLElement} element 
- * @param {Function} callback 
+ *
+ * @param {HTMLElement} element
+ * @param {Function} callback
  */
 export function callOnceWhenVisible(element, callback) {
   new IntersectionObserver((entries, observer) => {
--- a/src/js/histogram.js
+++ b/src/js/histogram.js
@@ -1,7 +1,7 @@
 import { Colors } from './colors';
 import debounce from './debounce';
 import { Metric } from './metric';
-import { el, prettyDate, chartExportOptions, drawMetricSummary } from './utils';
+import { el, prettyDate, chartExportOptions, drawMetricSummary, callOnceWhenVisible } from './utils';
 
 
 const [COLOR_DESKTOP, COLOR_MOBILE, COLOR_DESKTOP_ALT, COLOR_MOBILE_ALT] = Colors.getAll({rgba: true});
@@ -262,7 +262,10 @@ function drawHistogram(data, containerId, options) {
     });
   }
 
-  drawChart(series, containerId, options);
+  const chart = document.getElementById(`${options.metric}-chart`);
+  callOnceWhenVisible(chart, () => {
+    drawChart(series, containerId, options);
+  });
 };
 
 function drawChart(series, containerId, options) {
--- a/src/js/timeseries.js
+++ b/src/js/timeseries.js
@@ -2,7 +2,7 @@ import Changelog from './changelog';
 import { Colors } from './colors';
 import debounce from './debounce';
 import { Metric } from './metric';
-import { el, prettyDate, chartExportOptions, drawMetricSummary } from './utils';
+import { el, prettyDate, chartExportOptions, drawMetricSummary, callOnceWhenVisible } from './utils';
 
 
 function timeseries(metric, options, start, end) {
@@ -141,7 +141,10 @@ function drawTimeseries(data, options) {
     .then(flagSeries => series.push(flagSeries))
     // If the getFlagSeries request fails (503), catch so we can still draw the chart
     .catch(console.error)
-    .then(_ => drawChart(options, series));
+    .then(_ => {
+      const chart = document.getElementById(options.chartId);
+      callOnceWhenVisible(chart, () => drawChart(options, series));
+    });
 
 }
 let redrawTimeseriesTable = {};


src/js/utils.js
```diff
@@ -58,6 +58,28 @@ export const drawMetricSummary = (options, client, value, isMedian=true, change=
...
+ */
+export function callOnceWhenVisible(element, callback) {
+  new IntersectionObserver((entries, observer) => {
+    if (!entries[0].isIntersecting) {
+      return;
+    }
+
+    observer.unobserve(element);
+    callback();
+  }).observe(element);
```
tunetheweb, line 80: What about adding a threshold so charts are loaded when just off screen?

```suggestion
  }, {rootMargin: "1000px"}).observe(element);
```

rviscomi: That'd make it more likely for the first chart to load at startup and cause issues with early interactions. Not sure the UX is worth the INP tradeoff. WDYT?

tunetheweb: I was assuming that one chart wouldn't cause that much of an issue and it's the combination of doing so many charts at once that's the problem. Particularly for CrUX report which has 17 charts.

Personally I feel that lazy loading content (which effectively is what we're doing here - though it's more "lazy rendering") should be as invisible to the user as possible. Not rendering ANYTHING feels like we're optimising for the metric than the user. It's also the reason why `loading="lazy"` has thresholds to try to render some content as it approaches the viewport. So yeah I'd argue the UX is worth the INP trade off.

Plus, when the user clicks any chart, we scroll down and have to render ALL the graphs above it. At least getting the first one out of the way is one less to render in that case. Though, on the flip side, we'll likely render one more chart after the one we want too so there might not be any net gain, except when scrolling to the end (though the later one should at least render after the one being scrolled to).

We could try that and monitor INP scores? If this is enough to fix it then leave it in (or maybe even increase it!). If not, then can remove the threshold and go with 0.

src/js/utils.js
```diff
@@ -58,6 +58,28 @@ export const drawMetricSummary = (options, client, value, isMedian=true, change=
...
+/**
+ * Invokes the callback when the element is visible for the first time.
+ * 
+ * Rendering all charts immediately at startup causes long tasks and slow INP.
+ * Instead, wait until the chart is in view before rendering.
+ * Rendering a single chart might still create a long task,
+ * but it'll be less likely to contribute to INP.
+ * 
+ * @param {HTMLElement} element 
+ * @param {Function} callback
```
tunetheweb, line 70: Keep linter happy:

```suggestion
 *
 * Rendering all charts immediately at startup causes long tasks and slow INP.
 * Instead, wait until the chart is in view before rendering.
 * Rendering a single chart might still create a long task,
 * but it'll be less likely to contribute to INP.
 *
 * @param {HTMLElement} element
 * @param {Function} callback
```

tunetheweb (approved):
Happy to LGTM now tests are passing and will leave decision on whether to add a threshold to you. I'm more of the "we should add a threshold" side but at same time won't complain too much if we don't.

Good work!!

src/js/utils.js
rviscomi: Saw your LGTM comment, thanks.

> Not rendering ANYTHING feels like we're optimising for the metric than the user.

Just wanted to add that this is definitely my biggest concern with this approach. I think it's mitigated in part by the fact that rendering a single chart takes ~400ms (4x throttling on an M1), and IMO it's a more tolerable UX to wait that long to see the chart than it would be to wait that long after an interaction. And as a small consolation, rendering the chart when it's visible lets the user see the fancy animation of the data taking shape.

>Plus, when the user clicks any chart, we scroll down and have to render ALL the graphs above it. 

I tested this by jumping to a chart in the middle of the page and I'm only seeing that chart render, not the ones above it. Maybe that behavior is unique to IntersectionObserver, unlike `loading=lazy`?

Anyway, with your LGTM I'll resolve this and open a new PR to help with A/B testing the change.

tunetheweb: > I tested this by jumping to a chart in the middle of the page and I'm only seeing that chart render, not the ones above it.

Ohhhh it was only rendering all the charts above with my suggested change! So maybe that was a good reason not to do that!

PR #798, opened 2024-01-17
Add a param to analytics for A/B testing

rviscomi:
Progress on #796 

This change adds a `debug_version` param to the analytics with a value of `"defer-chart"`.

When releasing the recent INP improvements, we can split traffic 50/50 and use this param as a way of distinguishing the fix from the control.

I left a TODO in the code to remove the version when we're satisfied with the results and ramp traffic up to 100%. Alternatively, we could find a more permanent solution to annotate the website version (most recent last modified date?) in the version param.

### DIFF
--- a/src/js/send-web-vitals.js
+++ b/src/js/send-web-vitals.js
@@ -155,6 +155,9 @@ function sendWebVitals() {
       prefers_reduced_motion: prefersReducedMotion,
       prefers_color_scheme: prefersColorScheme,
       navigation_type: navigationType,
+
+      // TODO(rviscomi): Remove this after A/B testing the INP optimization.
+      debug_version: 'defer-chart',
     }, overrides);
 
     gtag('event', name, params);


tunetheweb:
I presume we can assume that a user will stay on a particular instance? i.e. they won't load the charts JS code from one instance (and use the lazy rendering), but then load web vitals from another? Imagine that's probably rare (if even possible)?

The other option would be to randomly set a cookie and then enable the new functionality (and analytics change) based on the presence/absence of that cookie. But cookie reading/writing could introduce it's own overheads.

Or set a random URL param on the reports menu and landing page and enable the new functionality (and analytics change) based on the presence/absence of that url param.

Both of those are probably overkill. Then again an A/B test may be overkill in and of itself given the low volume of traffic we get as could be quite noisy and skewed.

rviscomi:
I wouldn't expect it's possible for a single page load to use resources across A/B versions, especially given that we're cache-busting.

You can read more about the ways to split traffic in App Engine here: https://cloud.google.com/appengine/docs/legacy/standard/python/splitting-traffic#splitting_traffic_across_multiple_versions. I was planning to split by IP for simplicity.

Agreed that A/B might be overkill for our small site, but my hope is to create a case study from this and I think it sets a good example. If the data ends up being too noisy we can just roll it out and compare before/after.

tunetheweb (approved):

PR #816, opened 2024-02-21
Add an explicit warning about costs

rviscomi:
https://discuss.httparchive.org/t/warning-14-000-bigquery-charge-in-2-hours/2715

image


tunetheweb:
Think we should also add similar wording to https://github.com/HTTPArchive/httparchive.org/blob/main/docs/gettingstarted_bigquery.md

rviscomi:
Added

paulcalvano:
LGTM.

One thing we may want to add is a sentence that directs people to discuss.httparchive.org if they need help from the community on writing a query that is efficient.

PR #819, opened 2024-02-23
Remove desiredExecutionStart from LoAF attribution

### LINKED ISSUE #817: Investigate drop in LoAF attribution data
In the analytics for the httparchive.org website itself, the percentage of INP events having LoAF attribution data dropped at the end of last month:

image

```sql
SELECT
  event_date,
  COUNTIF(EXISTS(SELECT 1 FROM UNNEST(event_params) AS param WHERE STARTS_WITH(param.key, 'debug_loaf'))) AS nav_with_loaf,
  COUNT(0) AS nav_with_inp,
  COUNTIF(EXISTS(SELECT 1 FROM UNNEST(event_params) AS param WHERE STARTS_WITH(param.key, 'debug_loaf'))) / COUNT(0) AS pct_nav_with_loaf
FROM
  `httparchive.analytics_386559427.inp`
WHERE
  metric_name = 'INP' AND
  httparchive.analytics_386559427.FILTER_PREVIOUS_N_DAYS(event_date, 90, FALSE)
GROUP BY
  event_date
ORDER BY
  event_date
```

The lack of LoAF data makes it a lot harder to diagnose INP issues.

Possible issues:

- the API itself in Chrome (we're only logging for users with the experimental flag enabled)
- our origin trial registration
- our implementation/reporting
- our analysis

This could definitely be user error (me) but I wanted to file this issue as a reminder to investigate the reason.

FWIW I did take a quick look at the OT registration and it doesn't seem like it's expired or anything:

image

It will actually expire on March 5, and it's not renewable at this point. The LoAF API doesn't reach stable until later in the month. So there's going to be a bit of a blackout. Hopefully we can figure out what's going on and get some data flowing before the expiration.

rviscomi:
Fixes #817

### DIFF
--- a/src/js/send-web-vitals.js
+++ b/src/js/send-web-vitals.js
@@ -25,7 +25,7 @@ function sendWebVitals() {
       return entry.startTime < (loaf.startTime + loaf.duration) && loaf.startTime < (entry.startTime + entry.duration);
     }).forEach(loaf => {
       loaf.scripts.forEach(script => {
-        const totalDuration = script.startTime + script.duration - script.desiredExecutionStart;
+        const totalDuration = script.startTime + script.duration;
         if (totalDuration > loafAttribution.debug_loaf_script_total_duration) {
           loafAttribution = {
             // Stats for the LoAF entry itself.


tunetheweb (approved):
Oh good spot. Missed this in #787

PR #824, opened 2024-02-27
INP optimization and reporting improvements

rviscomi:
Table toggle (before):

image

After:

image

makeDatesPretty (before):

image

After:

image

Useful guide: https://webperf.tips/tip/date-formatting/

### DIFF
--- a/src/js/send-web-vitals.js
+++ b/src/js/send-web-vitals.js
@@ -25,33 +25,34 @@ function sendWebVitals() {
       return entry.startTime < (loaf.startTime + loaf.duration) && loaf.startTime < (entry.startTime + entry.duration);
     }).forEach(loaf => {
       loaf.scripts.forEach(script => {
-        const totalDuration = script.startTime + script.duration;
-        if (totalDuration > loafAttribution.debug_loaf_script_total_duration) {
-          loafAttribution = {
-            // Stats for the LoAF entry itself.
-            debug_loaf_entry_start_time: loaf.startTime,
-            debug_loaf_entry_end_time: loaf.startTime + loaf.duration,
-            debug_loaf_entry_work_duration: loaf.renderStart ? loaf.renderStart - loaf.startTime : loaf.duration,
-            debug_loaf_entry_render_duration: loaf.renderStart ? loaf.startTime + loaf.duration - loaf.renderStart : 0,
-            debug_loaf_entry_total_forced_style_and_layout_duration: loaf.scripts.reduce((sum, script) => sum + script.forcedStyleAndLayoutDuration, 0),
-            debug_loaf_entry_pre_layout_duration: loaf.styleAndLayoutStart ? loaf.styleAndLayoutStart - loaf.renderStart : 0,
-            debug_loaf_entry_style_and_layout_duration: loaf.styleAndLayoutStart ? loaf.startTime + loaf.duration - loaf.styleAndLayoutStart : 0,
+        const endTime = script.startTime + script.duration;
+        if (script.duration <= loafAttribution.debug_loaf_script_total_duration) {
+          return;
+        }
+        loafAttribution = {
+          // Stats for the LoAF entry itself.
+          debug_loaf_entry_start_time: loaf.startTime,
+          debug_loaf_entry_end_time: endTime,
+          debug_loaf_entry_work_duration: loaf.renderStart ? loaf.renderStart - loaf.startTime : loaf.duration,
+          debug_loaf_entry_render_duration: loaf.renderStart ? endTime - loaf.renderStart : 0,
+          debug_loaf_entry_total_forced_style_and_layout_duration: loaf.scripts.reduce((sum, script) => sum + script.forcedStyleAndLayoutDuration, 0),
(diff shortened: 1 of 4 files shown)


src/js/utils.js
```diff
@@ -22,12 +22,12 @@ export const chartExportOptions = {
...
         window.open(url, '_blank');
       },
-      text: 'Show Query'
+      text: 'Show query'
     }
   },
   buttons: {
     contextButton: {
-      menuItems: ['showQuery', 'downloadPNG']
+      menuItems: ['showQuery']
```
tunetheweb, line 30: What's this change about?

rviscomi: The "Download PNG image" button is extremely slow and it doesn't even work
image

src/js/send-web-vitals.js
```diff
@@ -25,33 +25,34 @@ function sendWebVitals() {
...
-            debug_loaf_meta_length: longAnimationFrames.length,
-          }
+        const endTime = script.startTime + script.duration;
+        if (script.duration <= loafAttribution.debug_loaf_script_total_duration) {
+          return;
+        }
+        loafAttribution = {
+          // Stats for the LoAF entry itself.
+          debug_loaf_entry_start_time: loaf.startTime,
+          debug_loaf_entry_end_time: endTime,
```
tunetheweb, line 35: These are not the same. Is this intentional?

src/js/send-web-vitals.js
```diff
@@ -25,33 +25,34 @@ function sendWebVitals() {
...
+          debug_loaf_entry_work_duration: loaf.renderStart ? loaf.renderStart - loaf.startTime : loaf.duration,
+          debug_loaf_entry_render_duration: loaf.renderStart ? endTime - loaf.renderStart : 0,
```
tunetheweb, line 37: Same here

src/js/send-web-vitals.js
```diff
@@ -25,33 +25,34 @@ function sendWebVitals() {
...
+          debug_loaf_entry_total_forced_style_and_layout_duration: loaf.scripts.reduce((sum, script) => sum + script.forcedStyleAndLayoutDuration, 0),
+          debug_loaf_entry_pre_layout_duration: loaf.styleAndLayoutStart ? loaf.styleAndLayoutStart - loaf.renderStart : 0,
+          debug_loaf_entry_style_and_layout_duration: loaf.styleAndLayoutStart ? endTime - loaf.styleAndLayoutStart : 0,
```
tunetheweb, line 40: And here

src/js/send-web-vitals.js
```diff
@@ -25,33 +25,34 @@ function sendWebVitals() {
...
+
+          // Stats for the longest script in the LoAF entry.
+          debug_loaf_script_total_duration: script.duration,
```
tunetheweb, line 43: This looks better as it's now a duration, but it is a change as previously `totalDuration` was the endTime.

tunetheweb (commented):
Have some questions on the send-web-vitals.js changes but maybe they are intentional?

src/js/send-web-vitals.js
rviscomi: Yeah we should have been reporting the duration, not the end time, so this is a fix. In the conditional above, we were also only looking for the latest script instead of the longest one.

src/js/send-web-vitals.js
```diff
@@ -24,34 +24,35 @@ function sendWebVitals() {
...
+          debug_loaf_entry_end_time: loafEndTime,
+          debug_loaf_entry_work_duration: loaf.renderStart ? loaf.renderStart - loaf.startTime : loaf.duration,
+          debug_loaf_entry_render_duration: loaf.renderStart ? loafEndTime - loaf.renderStart : 0,
+          debug_loaf_entry_total_forced_style_and_layout_duration: loaf.scripts.reduce((sum, script) => sum + script.forcedStyleAndLayoutDuration, 0),
+          debug_loaf_entry_pre_layout_duration: loaf.styleAndLayoutStart ? loaf.styleAndLayoutStart - loaf.renderStart : 0,
+          debug_loaf_entry_style_and_layout_duration: loaf.styleAndLayoutStart ? loafEndTime - loaf.styleAndLayoutStart : 0,
+
+          // Stats for the longest script in the LoAF entry.
+          debug_loaf_script_total_duration: script.duration,
+          debug_loaf_script_compile_duration: script.executionStart - script.startTime,
```
tunetheweb, line 44: Nit.

It's not super obvious these are the longest script? Although I get it's singular ("script", rather than "scripts"), I think even the singular can be taken to mean "all [Java]Script code".

In particular I would read "debug_loaf_script_total_duration" as the total duration of ALL LoAF scripts.

But this is not a chance in this PR and probably a bit late to raise it now.

tunetheweb (approved):
One nit comment but probably best to leave as is.

tunetheweb:
BTW can we update Web Almanac as well? Would like to keep them in sync, even though I know we're not looking at optimising that one (yet!).

src/js/send-web-vitals.js
rviscomi: Yeah this is good feedback for a future web-vitals implementation

rviscomi:
>BTW can we update Web Almanac as well? Would like to keep them in sync, even though I know we're not looking at optimising that one (yet!).

Will do

Issue #826, opened 2024-02-29
Update "Passes CWV" metric to use INP


rviscomi:
With INP replacing FID on March 12, we need to update the [Passes CWV](https://httparchive.org/reports/chrome-ux-report#cruxPassesCWV) metric in the CrUX report to use INP instead.

PR #829, opened 2024-03-05
Update CWV report to replace FID with INP

rviscomi:
Progress on #826 

- Updates the "Passes CWV" metric description to refer to INP as one of the three CWV metrics
- Updates URLs to Web Vitals docs to skip a redirect
- Adds a note about FID being used prior to March 2024
- Rearranges INP and FID metrics in the report so that INP appears first

### DIFF
--- a/config/reports.json
+++ b/config/reports.json
@@ -566,7 +566,7 @@
     "cruxPassesCWV": {
       "name": "Passes Core Web Vitals",
       "type": "%",
-      "description": "The percentage of origins passing all three [Core Web Vitals](https://web.dev/vitals/#core-web-vitals) (LCP, FID, CLS) with a \"good\" experience. Note that if an origin is missing FID data, it's assessed based on the performance of the remaining metrics.",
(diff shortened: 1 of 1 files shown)


config/reports.json
```diff
@@ -482,7 +482,7 @@
       "name": "Cumulative Layout Shift",
       "type": "",
       "singular": "",
-      "description": "The layout shift scores for every unexpected layout shift that occurs during the entire lifespan of the page. See the [Cumulative Layout Shift (CLS)](https://web.dev/cls/) for more info.",
+      "description": "The layout shift scores for every unexpected layout shift that occurs during the entire lifespan of the page. See the [Cumulative Layout Shift (CLS)](https://web.dev/articles/cls/) for more info.",
```
tunetheweb, line 485: Can we remove final slash from all of these while we're at it?

```suggestion
      "description": "The layout shift scores for every unexpected layout shift that occurs during the entire lifespan of the page. See the [Cumulative Layout Shift (CLS)](https://web.dev/articles/cls) for more info.",
```

rviscomi: What exactly does this optimize? 😅

tunetheweb: Google Analytics, so it doesn't track it as two different pages, for example:

image

We probably should also update it to strip off the final `/`.

rviscomi: Ok SG

tunetheweb (approved):
LGTM

Issue #854, opened 2024-05-08
Update beta header


rviscomi:
<img width="411" alt="image" src="https://github.com/HTTPArchive/httparchive.org/assets/1120896/b9d2936c-efbd-4637-82e4-edaba7fe11d1">

- Change "Beta" styles to appear less like a button
- Add link to GitHub to file an issue: https://github.com/HTTPArchive/httparchive.org/issues

rviscomi:
Also update the "Feedback" section in the footer to point to this repo for any feedback/issues

<img width="1196" alt="image" src="https://github.com/HTTPArchive/httparchive.org/assets/1120896/e057b6dc-ad33-479c-a5af-25459e10a12f">

Issue #853, opened 2024-05-02
Improve tech report page weight formatting


rviscomi:
<img width="1166" alt="image" src="https://github.com/HTTPArchive/httparchive.org/assets/1120896/8da61f0c-4852-43ec-97a7-409eb8f7aaca">

In this screenshot the image weight is shown as "752149 kB"

This seems to be showing the number of bytes, not KB. Let's convert from bytes to KB and if the value is still larger than 1000, convert again from KB to MB. Round to the nearest whole value to avoid showing any decimal places.

Also, as in the existing [Page Weight report](https://httparchive.org/reports/page-weight#bytesTotal) we typically format the unit as capital-K "KB":

<img width="520" alt="image" src="https://github.com/HTTPArchive/httparchive.org/assets/1120896/2ac5a8e3-552e-4573-90ae-36c2480e3d45">

tunetheweb:
Relevant: https://github.com/HTTPArchive/httparchive.org/issues/643

Issue #851, opened 2024-05-02
Beta Report - Confusion about where the data is


jordyscholing:
I didn't know where to look after I changed the field and pressed "update". Since it is still in beta I didn't think much of it as data might not work yet. Turned out I had to scroll down. Maybe indicate or somehow make it clearer that the data is below the fold.
![image](https://github.com/HTTPArchive/httparchive.org/assets/163306795/1dff015a-43fb-4137-b3df-fc9c325e0021)

This is especially something for the "comparison" tab, as there are more input fields and you don't see that there's something below the fold directly. On the "drilldown" tab you have some kind of indication: 
![image](https://github.com/HTTPArchive/httparchive.org/assets/163306795/a93eea01-478b-4961-94b2-69277ff10121)

tunetheweb:
Maybe we should scroll down to the chart after clicking Update button?

jordyscholing:
> Maybe we should scroll down to the chart after clicking Update button?

Yeh, sounds good to me :)

Issue #850, opened 2024-05-02
Beta Report - Blank Dropdown List Items in Dark Theme


Sylvhub:
When I switch to the _Dark Theme_ and click on the 'Technology' dropdown list, the only item visible is the one I select; the rest of the list appears blank. Please refer to the attached screenshot for more details.
![image](https://github.com/HTTPArchive/httparchive.org/assets/17831400/84706d39-4ede-4942-829c-c80ee8f90105)

PR #858, opened 2024-05-16
Bugfixes and minor improvements

sarahfossheim:
## Summary of the changes
Changed from reported bugfixes:
- After submitting a new search, the browser scrolls to the main content
Solves issue: https://github.com/HTTPArchive/httparchive.org/issues/851
- Page weight in the graph is displayed in bytes (label corrected)
Solves issue: https://github.com/HTTPArchive/httparchive.org/issues/853
- Page weight in the summary view is either bytes, KB, or MB
Solives issue: https://github.com/HTTPArchive/httparchive.org/issues/853
- Date is no longer converted to local timezone
Solves issue: https://github.com/HTTPArchive/httparchive.org/issues/849
- Change "beta" label styling + link to this repo
Solves issue: https://github.com/HTTPArchive/httparchive.org/issues/854
- Give the options in the dropdown a dark font color in dark theme, to avoid white-on-white text in Edge
Solves issue: https://github.com/HTTPArchive/httparchive.org/issues/850
- Change dev to prod API
Solves issue: https://github.com/HTTPArchive/httparchive.org/issues/864

Additional changes:
- Removing image placeholders
- Hovers on the chart now also show corresponding symbol, and combine all data for the same date (eg both desktop and mobile)
- Two different types of values in the pageweight summary within the timeseries: in bytes (as displayed on the line chart) + formatted (as used elsewhere)

## Screenshots
Timezone San Francisco - main branch
Hover on a chart showing data for Jan 31st

Timezone San Francisco - changes
Hover on a chart showing data for Feb 1

### DIFF
--- a/src/js/techreport/timeseries.js
+++ b/src/js/techreport/timeseries.js
@@ -150,10 +150,6 @@ class Timeseries {
           valueLabel.textContent = categoryData?.[breakdown.name]?.[summary];
           valueLabel.classList.add('breakdown-value');
           itemWrapper.appendChild(valueLabel);
-
-          const summaryLabel = document.createElement('p');
-          summaryLabel.textContent = `${latestValue}${breakdown.suffix || ''}`;
-          itemWrapper.appendChild(summaryLabel);
         } else {
           /* Add the value to the wrapper */
           const valueLabel = document.createElement('p');
--- a/src/js/techreport/timeseries.js
+++ b/src/js/techreport/timeseries.js
@@ -205,14 +205,12 @@ class Timeseries {
         const timestamp = component.querySelector('[data-slot="timestamp"]');
         const label = card.getElementsByClassName('breakdown-label')[0];
         const value = card.getElementsByClassName('breakdown-value')[0];
-        const secondaryValue = card.getElementsByClassName('breakdown-value-secondary')[0];
 
         /* Update text */
         label.innerHTML = latest.technology;
         if(latestValue) {
           if(summary) {
             value.innerHTML = `${summaryValue}`;
-            secondaryValue.innerHTML = `${latestValue}${config.series.suffix || ''}`;
           } else {
             value.innerHTML = `${latestValue}${config.series.suffix || ''}`;
           }
--- a/src/js/techreport/timeseries.js
+++ b/src/js/techreport/timeseries.js
@@ -108,6 +108,7 @@ class Timeseries {
     /* Get settings */
     const metric = viz.dataset.metric;
     const endpoint = viz.dataset.endpoint;
+    const summary = viz.dataset.summary;
 
     const app = pageFilters.app[0];
     const filtered = data?.[app]?.filter(entry => entry[endpoint]);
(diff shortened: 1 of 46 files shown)


src/js/techreport/timeseries.js
```diff
@@ -184,17 +198,24 @@ class Timeseries {
...
-          value.innerHTML = `${latestValue}${config.series.suffix || ''}`;
+          if(summary) {
+            value.innerHTML = `${summaryValue}`;
+            secondaryValue.innerHTML = `${latestValue}${config.series.suffix || ''}`;
+          } else {
+            value.innerHTML = `${latestValue}${config.series.suffix || ''}`;
+          }
         } else {
           value.classList.add('undefined');
           value.innerHTML = 'No data';
```
tunetheweb, line 221: Nit but we've recently discovered `innerHTML` (and `innerText`) is actually a bit slower than `textContent`.

It if really is just text (and I'm not sure if it always is), could we use `textContent`?

tunetheweb (approved):
LGTM with one nit

src/js/techreport/utils/data.js
```diff
@@ -62,11 +62,13 @@ const parsePageWeightData = (metric, date) => {
       ...submetric,
       desktop: {
         ...submetric.desktop,
+        median_bytes_formatted: submetric.desktop.median_bytes > 1024 ? `${Math.round(submetric.desktop.median_bytes / 1024)} KB` : submetric.desktop.median_bytes > 1048576 ? `${Math.round(submetric.desktop.median_bytes / 1048576)} MB` : `${submetric.desktop.median_bytes} bytes`,
```
rviscomi, line 65: This will always show KB even when it's > 1 MB

```suggestion
        median_bytes_formatted: submetric.desktop.median_bytes > 1048576 ? `${Math.round(submetric.desktop.median_bytes / 1048576)} MB` : submetric.desktop.median_bytes > 1024 ? `${Math.round(submetric.desktop.median_bytes / 1024)} KB` : `${submetric.desktop.median_bytes} bytes`,
```

src/js/techreport/utils/data.js
```diff
@@ -62,11 +62,13 @@ const parsePageWeightData = (metric, date) => {
...
         client: 'desktop',
         date: date,
       },
       mobile: {
         ...submetric.mobile,
+        median_bytes_formatted: submetric.mobile.median_bytes > 1024 ? `${Math.round(submetric.mobile.median_bytes / 1024)} KB` : submetric.mobile.median_bytes > 1048576 ? `${Math.round(submetric.mobile.median_bytes / 1048576)} MB` : `${submetric.mobile.median_bytes} bytes`,
```
rviscomi, line 71: Same comment. (should this conversion be moved to a reusable function?)

templates/techreport/components/timeseries.html
```diff
@@ -39,13 +39,15 @@ <h4>
               <div class="breakdown-item">
                 <p class="breakdown-label">{{ breakdown.name }}</p>
                 <p class="breakdown-value">00{% if breakdown.suffix == "%" %}.00{% endif %}{{ breakdown.suffix }}</p>
+                <p class="breakdown-value-secondary"></p>
```
rviscomi, line 42: Is this only needed because the chart is always in bytes? TBH I don't love the redundancy and I think we can get away with only showing the formatted value. Alternatively, WDYT about stuffing the # of bytes into the `title` attribute of the formatted value?

image

src/js/techreport/timeseries.js
sarahfossheim: So far it's only text, can update to `textContent`

src/js/techreport/utils/data.js
sarahfossheim: Good point, will move to a function in the utils 👍🏻

sarahfossheim:
@rviscomi updated :)

templates/techreport/components/timeseries.html
sarahfossheim: I removed it for now since the `title` comes with accessibility challenges and is easily overlooked anyway. If people need to hover to see the value they can just as well hover over the graph (which works with keyboard etc) 😄

If we get feedback that it's confusing that one is in MB/KB and the other one in bytes, then maybe we can add a custom (accessible) tooltip on the formatted values.

rviscomi (approved):
LGTM thanks!

PR #876, opened 2024-06-10
Add GA4 fields to match documentation

tunetheweb:
Woudl like to match docs: https://github.com/GoogleChrome/web-vitals#send-the-results-to-google-analytics

Which would also allow these queries to work without having to alter: https://web.dev/articles/vitals-ga4

### DIFF
--- a/src/js/send-web-vitals.js
+++ b/src/js/send-web-vitals.js
@@ -54,7 +54,7 @@ function sendWebVitals() {
     }));
   }
 
-  function sendWebVitalsGAEvents({name, delta, id, attribution, navigationType}) {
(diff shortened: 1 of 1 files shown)


rviscomi (approved):

rviscomi:
Nit: updated the title to reflect that no fields were changed, only new ones added. Changing existing field names would require an update to the BQ/Looker pipeline.

tunetheweb:
> Nit: updated the title to reflect that no fields were changed, only new ones added. Changing existing field names would require an update to the BQ/Looker pipeline.

And that's exactly why I didn't remove them :-)

But fair point on the title!

PR #909, opened 2024-08-08
RUM improvements

rviscomi:
FID is deprecated and we're not actively monitoring it, so let's remove it from our analytics pipeline.

This fixes an unnecessary limitation in our LoAF attribution that required LoAFs to have script entries. Some data is better than none, so we will beacon whatever is available.

This also improves CSP for SpeedCurve and GA:

image


rviscomi:
The directives in the [GA4 CSP docs](https://developers.google.com/tag-platform/security/guides/csp#google_analytics_4_google_analytics) don't include these hosts, but they do seem to be used (if not necessary):

- analytics.google.com
- stats.g.doubleclick.net
- td.doubleclick.net

rviscomi:
Turns out we actually had Google Signals enabled in our analytics config, which was responsible for the additional hosts. I've disabled it, so we shouldn't need them in our policy anymore.

PR #934, opened 2024-09-24
Category dashboard with collapsible filters

sarahfossheim:
## New features
* New compact header only used on the tech report pages
* Sidebar with filters, instead of big card on top of the page
* Prototype of the categories page:
  * Category level summary: number of origins and number of technologies within a category
  * Table with summarising info on all techs within that category
  * Possibility to click through to a specific tech, or compare multiple (or all) technologies within the tech
  * Prototype sortable table functionality _(not in the UI yet)_
  * `/reports/techreport/category`
* Only one view for both drilldown and comparison view 
  * Add more technologies to compare in the sidebar
  * New template loads depending on the amount of techs selected
  * See `/reports/techreport/tech`

## (Visual) Improvements
* Made clickable area in the landing page cards clickable (https://github.com/HTTPArchive/httparchive.org/issues/990)
* Responsive menu, turning into a dropdown on mobile (+ updating `aria-expanded`)
* When there are multiple technologies in the filter, they become numbered (same for the remove button alt texts)
* Comparison table now shows Overall Good CWVs as a percentage (https://github.com/HTTPArchive/httparchive.org/issues/991)
* Removed or replaced placeholder texts (https://github.com/HTTPArchive/httparchive.org/issues/995)
* Limited amount of categories displayed in the category description (https://github.com/HTTPArchive/httparchive.org/issues/978)

## Bugfixes
* Updates to the responsiveness/zoom _(more improvements to come in future PR)_
* The placeholder bright yellow color used for debugging WordPress line changed to color with enough contrast
* Updated copywriting for some section _(more improvements to come in future PR)_
* Fixed keyboard order for the points on comparison graphs (it was reversed)
* Geo & Rank are remembered between categories and technology dashboard
* Sort table alternatives from recent to old (https://github.com/HTTPArchive/httparchive.org/issues/996)

---

#### One technology

Technology drill down view with 'All technologies' selected from the filter sidebar. The page shows 4 summary cards (latest CWVs, LH performance, LH accessibility, and page weight)

#### Multiple technologies

Desktop

Desktop version of the comparison view: filter sidebar on the left showing all techs, wordpress, wix, eleventy, and a comparison table on the right. The nav in the header has items: HttpArchive logo, all reports, tech report, technologies, categories

Mobile

Mobile version of the comparison view: the site navigation is hidden behind a hamburger icon, the filters are collapsed into a top navigation bar as well

### Categories

Cateogries page with CMS selected from the filters. There's a card showing # of origins and # of technologies within the category, and underneath there's a summary table


src/js/techreport/tableLinked.js
```diff
@@ -21,18 +25,56 @@ class TableLinked {
...
+      const geo = filters.get('geo') || 'ALL';
+      const rank = filters.get('rank') || 'ALL';
+
+      // sort data
+      const sortEndpoint = component.dataset.sortEndpoint;
+      const sortMetric = component.dataset.sortMetric;
+      const sortKey = component.dataset.sortKey;
+      const client = component.dataset.client;
+
+      console.log(sortMetric);
```
rviscomi, line 47:
```suggestion
```

src/js/techreport/tableLinked.js
```diff
@@ -21,18 +25,56 @@ class TableLinked {
...
+          const bMetric = bLatest[sortEndpoint].find(row => row.name === sortMetric);
+
+          const aValue = aMetric[client][sortKey];
+          const bValue = bMetric[client][sortKey];
+
+          return bValue - aValue > 0 ? 1 : -1;
+        });
+      }
+
+      this.dataArray.forEach(technology => {
```
rviscomi, line 67: I know we talked about pagination before, but I don't remember if we talked about performance improvements to the initial API requests.

So for example if we only want to show the top 10 technologies by default and we can assume that the categories API endpoint lists the technologies in descending popularity order, we could make requests to the other API endpoints with only those first 10 technologies as request params to improve performance a bit.

config/techreport.json
```diff
@@ -1118,6 +1124,89 @@
...
+                "hiddenName": true,
+                "type": "checkbox"
+              },
+              {
+                "key": "technology",
+                "name": "Tech",
+                "type": "heading"
+              },
+              {
+                "key": "origins",
```
rviscomi, line 1157: In addition to the number of origins, I think a primary use case for this data would be to see the proportion of origins that adopt this category relative to all origins that adopt this category.

For example, there are 3.2M WordPress origins and the /categories endpoint says that there are 8.8M origins in the CMS category, so we could show that "market share" value as 36%.

What do you think would be the best way to show the 36% stat? It could be a new column, combined with the # of origins somehow, or replace the # origin column entirely. Alternatively, would you hate a pie chart at the top of the page?

_Aside: I suspect the 8.8M number is wrong as it's double-counting origins that use multiple CMSs. WordPress market share should be more like 75% of the CMS category. This is something we'll need to fix in BigQuery._

cc @max-ostapenko

templates/techreport/category.html
```diff
@@ -0,0 +1,117 @@
...
+  <div class="report-content">
+    <div class="info-panel">
+      <div class="block-l">
+        <p class="info-label">Beta</p>
+        <p>We're still actively working on the categories page.</p>
+      </div>
+    </div>
+
+    <div class="intro block-m">
+      <div class="info">
```
rviscomi, line 88: When the API supports it, we should also have a placeholder for the category description here.

rviscomi, line 88: The /category endpoint returns the total number of origins in the category, which I think could be useful summary info here at the top. Just not sure how it should look.

src/js/techreport/tableLinked.js
```diff
@@ -41,10 +83,41 @@ class TableLinked {
...
-              link.setAttribute('href', `?tech=${app}`);
-              const formattedApp = DataUtils.formatAppName(app);
-              link.textContent = formattedApp;
+              link.setAttribute('href', `/reports/techreport/tech?tech=${app}&geo=${geo}&rank=${rank}`);
+              link.innerHTML = formattedApp;
               cell.append(link);
+            } else if (column.type === 'checkbox') {
+              cell = document.createElement('td');
+              const formattedApp = DataUtils.formatAppName(app);
+              const checkbox = document.createElement('input');
```
rviscomi, line 92: Could it be useful to have a "(de)select all" checkbox in the header?

config/techreport.json
rviscomi, line 1157: When filtering to the top 1k rank, I'm seeing technologies with "null" origin counts

https://technology-report-dot-httparchive.uk.r.appspot.com/reports/techreport/category?tech=&geo=ALL&rank=Top+1k&category=CMS

image

src/js/techreport/tableLinked.js
```diff
@@ -21,18 +25,56 @@ class TableLinked {
...
+      const client = component.dataset.client;
+
+      console.log(sortMetric);
+      if(sortMetric) {
+        this.dataArray = this.dataArray.sort((techA, techB) => {
+          // Sort techs by date to get the latest
+          const aSortedDate = techA.sort((a, b) => new Date(b.date) - new Date(a.date));
+          const bSortedDate = techB.sort((a, b) => new Date(b.date) - new Date(a.date));
+          const aLatest = aSortedDate[0];
+          const bLatest = bSortedDate[0];
```
rviscomi, line 54: "Latest" should be the same for all technologies. It's possible for a technology to exist at one point in the dataset and then disappear forever, but this would always compare that ancient technology against the latest versions of all other modern technologies.

For example, October CMS shows up at the top of the top 1k CMS list with 361 origins:

https://technology-report-dot-httparchive.uk.r.appspot.com/reports/techreport/category?tech=&geo=ALL&rank=Top+1k&category=CMS

image


But in the technology drilldown it's clearer that the most recent (and only) data point is from April 2022:

https://technology-report-dot-httparchive.uk.r.appspot.com/reports/techreport/tech?tech=October%20CMS&geo=ALL&rank=Top%201k

image

The expected behavior is for it to not appear on the Category page at all since it's not part of the latest dataset. That might also be a good reason to explicitly have a label like `Latest data: 2024-11-01` at the top of this report, similar to the others.

templates/techreport/templates/selector_tech.html
```diff
@@ -1,18 +1,18 @@
 <div class="tech-selector-group" data-tech="{{ name }}">
   <div class="content">
-    <div class="categories-selector-group">
+    <!-- <div class="categories-selector-group">
```
rviscomi, line 3: This HTML is commented out. Should it be removed?

templates/techreport/category.html
```diff
@@ -0,0 +1,117 @@
...
+<form id="page-filters">
+  <div>
+    <fieldset class="technology-filters">
+      <legend class="form-title">Choose a dataset</legend>
+
+      <div class="tech-selector">
+        <label for="categories_selector" class="tech">Category</label>
+        <div class="tech-input-wrapper">
+          {% set category_selected = tech_report_page.filters.category or tech_report_config.default_category %}
+          <select name="categories" class="tech" data-selected="{{ category_selected }}" id="categories_selector">
```
rviscomi, line 13: I'm finding the native` <select> `hard to use for such long lists of options. Do you have any thoughts on using a [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) instead? I like having the ability to start typing what I'm looking for and it can match any position in the value, whereas typing in a `<select>` only works if it's a prefix match, and the datalist also filters down the results to make it easier to find what you're looking for.

Here's a production example: https://chromestatus.com/metrics/feature/popularity

image

Maybe something to consider for the technologies inputs too.

templates/techreport/category.html
```diff
@@ -0,0 +1,117 @@
...
+      {% set client = request.args.get('client', '')  or 'mobile' %}
+      {% set sort_endpoint = "adoption" %}
+      {% set sort_metric = "adoption" %}
+      {% set sort_key = "origins" %}
+      {% set sort_order = "desc" %}
+
+      {% set table = component.table %}
+      {% include "techreport/components/table_linked.html" %}
+
+      <a data-name="selected-apps" href="/reports/techreport/tech" class="cta-link">Compare technologies</a>
```
rviscomi, line 114: If there are geo/rank/client filters set on the Categories page and I click this to compare multiple technologies, it looks like those settings are lost on the comparison page.

USA, Top 100k, Desktop:

image

All, All, mobile:

image

rviscomi (commented):
Staged here and left a bunch of feedback: https://technology-report-dot-httparchive.uk.r.appspot.com/reports/techreport/category?geo=ALL&rank=ALL&category=CMS

I had to patch one bug to get it to work (the fix is in one of the comment suggestions) but otherwise these are all just feature requests. Great work!

templates/techreport/category.html
sarahfossheim: Yes, agree some sort of text input is needed for those long dropdowns. I suggest doing it in a follow-up PR, so this one doesn't become even bigger to review? Should I create an issue for it?

templates/techreport/category.html
sarahfossheim: Working on it, I suggest summary cards similar to the `/drilldown` page
two cards with 'Origins' and 'Technologies' containing placeholder data, on top of a table with technologies within a category

rviscomi: LGTM!

templates/techreport/category.html
rviscomi: Sure, thanks

config/techreport.json
sarahfossheim: > What do you think would be the best way to show the 36% stat? It could be a new column, combined with the # of origins somehow, or replace the # origin column entirely.

If a primary use case is to see the market share, I'm inclined to say replace the origins with the market share, so the table doesn't become overcrowded with info. The other numbers in the table (% of good CWV etc) feel more important than the absolute number of origins to provide a summary of the category data as well, so if a column has to be hidden to make space for the market share, I think origins is the safest one to remove.

> Alternatively, would you hate a pie chart at the top of the page?

What about just for the top 5/top 10 technologies within a category? I fear anything more than a handful of items will become too busy/hard to read, at which point it's not very useful anymore and the table becomes a better alternative.

> Aside: I suspect the 8.8M number is wrong as it's double-counting origins that use multiple CMSs

If some origins can have multiple CMSes, then adding all the market share percentages can end up being more than 100%, and a pie chart doesn't work anymore? So if that's the case, a different way of highlighting the top techs within a category might be better?

→ Sounds like marketshare might be better for a follow-up issue & PR as well, so we can think through the specifics of it a bit more?

rviscomi: SGTM

src/js/techreport/tableLinked.js
```diff
@@ -21,18 +25,56 @@ class TableLinked {
...
+
+          // Get the correct endpoint & metric
+          const aMetric = aLatest[sortEndpoint].find(row => row.name === sortMetric);
```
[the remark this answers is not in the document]
sarahfossheim: That's updated now 👍🏻

templates/techreport/templates/selector_tech.html
sarahfossheim: That's updated now 👍🏻

templates/techreport/category.html
sarahfossheim: There was indeed an issue with the saved filters, should be fixed now 👍🏻

templates/techreport/category.html
sarahfossheim: https://github.com/HTTPArchive/httparchive.org/issues/993

config/techreport.json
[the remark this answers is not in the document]
sarahfossheim: Formatted them with commas everywhere now

src/js/techreport/tableLinked.js
sarahfossheim: Created a follow-up issue for the pagination: https://github.com/HTTPArchive/httparchive.org/issues/997

sarahfossheim:
@rviscomi - I made a lot of updates based on the feedback you left, and created follow-up action items for the larger todos (sorting of the table f.ex) so those can go in a follow-up PR. Also fixed a few other things at the same time, like removing/replacing the placeholder texts and making the navigation responsive (hamburger menu). 

So it's ready for re-review :)

tunetheweb (approved):
OK I think this is good to merge!

PR #948, opened 2024-11-08
Docs cleanup

max-ostapenko:
Cleanup related to https://github.com/HTTPArchive/har.fyi/pull/15 + a quick link fix in FAQ.


tunetheweb (changes requested):
Maybe rather than removing completely we should replace these with links to the new docs? Since we can't do redirects for GitHub.

max-ostapenko:
That's true, restored the `Getting started`.
[Repo insights](https://github.com/HTTPArchive/httparchive.org/graphs/traffic) show 35 users.

The other docs were moved somewhere in May, and I replaced all the referring links then.
Should be fine to remove today.

tunetheweb (approved):

PR #962, opened 2024-11-29
BQ docs link update

max-ostapenko:
Why do we bring users to a small paragraph in the FAQ, while we could take them directly to the docs?

And the answer duplicates the information from the guide a lot.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -36,13 +36,9 @@ Given these conditions it's virtually impossible to compare the HTTP Archive's t
 
 ## How do I use BigQuery to write custom queries over the data?
 
-The HTTP Archive dataset is available on BigQuery. Be aware that as a consequence of collecting so much metadata from millions of websites each month, the dataset is _extremely large_—multiple petabytes. Care _must_ be taken to set up [cost controls](https://cloud.google.com/bigquery/docs/custom-quotas) to avoid unexpected bills. Also see our guide to [minimizing query costs](https://har.fyi/guides/minimizing-costs/) for tips on staying under the 1 TB per month free quota.
-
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -36,14 +36,10 @@ Given these conditions it's virtually impossible to compare the HTTP Archive's t
...
-
-Check out [Getting Started Accessing the HTTP Archive with BigQuery](https://har.fyi/guides/getting-started/), a guide for first-time users written by [Paul Calvano](https://twitter.com/paulcalvano).
+The HTTP Archive dataset is available publicly on BigQuery. Check out [Getting Started Accessing the HTTP Archive with BigQuery](https://har.fyi/guides/getting-started/).
 
 For a guided walkthrough of the project, watch this in-depth [30 minute video](https://www.youtube.com/watch?v=00f9kza3BJ0) featuring HTTP Archive maintainer [Rick Viscomi](https://twitter.com/rick_viscomi).
 
-If you have any questions about using BigQuery, reach out to the HTTP Archive community at [discuss.httparchive.org](https://discuss.httparchive.org/).
```
tunetheweb, line 45: Maybe keep this section?

tunetheweb (approved):
Yeah I was in two minds about this myself when I added the menu link.

docs/faq.md
max-ostapenko: We could check the clicks in analytics.
IMHO the forum is visible enough: header + footer on httparchive.org + several pages in the guide.

BTW I'll rather add it [to the header links](https://github.com/HTTPArchive/har.fyi/blob/66cafadc7f2efea5d842197c46db25868ce61024/astro.config.mjs#L28).

tunetheweb: Yeah that's fair enough. BTW let's add Bluesky to the har.fyi since many people are against Twitter now.

PR #1007, opened 2025-02-14
[WIP] Limit amounts of techs shown per page

sarahfossheim:
⚠️ Work in progress PR, not ready for review/merging yet

---

**Issue**: Currently we can select an infinite amount of techs on the comparison page, and some categories have ca 300 technologies in them as well. When rendering the data for this many technologies, the page ends up loading slowly (and the visualizations hard to read).

**This PR will add**:
- Pagination for the categories page, showing only 10 technologies per page
- Limit to how many technologies can manually be selected in the comparison view
- Other performance improvements, like only fetching tech and category names for the dropdowns (currently big response)


max-ostapenko:
Strange filters behaviour, recording:
https://github.com/user-attachments/assets/536070c6-7a71-47b8-999f-9f751779164f

sarahfossheim:
@tunetheweb Will add point 1 (selecting number of techs per page) to this PR as well 👍🏻 
What kind of filtering do you mean for point 2?  

@max-ostapenko should be fixed now :)

tunetheweb:
> @tunetheweb Will add point 1 (selecting number of techs per page) to this PR as well 👍🏻

SGTM

> What kind of filtering do you mean for point 2?

Say I run a CMS called BarryPress. I wanna compare it to the top CMS's. So I go to CMS category, and I select a few of the top ones that I wanna compare mine too (WordPress, Shopify, Wix, Joomla) and then wanna add mine I have to paginate through all the technologies until I find BarryPress (which is way down the list). PReviously when we listed them all I could at least Ctrl+F to find BarryPress on the current page, now I can't.

So having some sort of type ahead filter might make it easy for me to find and add BarryPress to the comparison.

Alternatively, I could just click those top 4, and then add BarryPress on the next screen with the "Add another Technology" button. So there is a way to do it, which is probably fine for now as imagine the typeahead filter could be quite a bit of work.

max-ostapenko:
@sarahfossheim  Somehow I get a long list of technologies again instead of pagination.

Regarding filtering, I would suggest the user to use the technologies page for a scenario like this.

sarahfossheim:
@tunetheweb ready for review :) 

Linter complains about `</option>`s without start tag, but they do have a start tag. Not sure what's up with that 🤔 Because of the if statement?

```
<select id="rowsPerPage">
    <option value="10" {% if filters.rows == "10" %}selected{% endif %}>10</option>
    <option value="25" {% if filters.rows == "25" %}selected{% endif %}>25</option>
    <option value="50" {% if filters.rows == "50" %}selected{% endif %}>50</option>
</select>
```

tunetheweb:
I've hit this in the past. Think I had to do this to get round it:

```jinja
<select id="rowsPerPage">
    {% if filters.rows == "10" %}
    <option value="10" selected>10</option>
    {% else %}
    <option value="10">10</option>
    {% endif %}
    {% if filters.rows == "25" %}
    <option value="25" selected>25</option>
    {% else %}
    <option value="25">25</option>
    {% endif %}
    {% if filters.rows == "50" %}
    <option value="50" selected>50</option>
    {% else %}
    <option value="50">50</option>
    {% endif %}
</select>
```

Which is a bit yucky, but was the only way I found to stop the linter complaining :-(

tunetheweb:
Looks like CI is failing for other things:
- More CLS. Think you said you were working on this?
- Console error when looking at http://127.0.0.1:8080/reports/techreport/comparison

```
section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1 Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at new i (section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1:3984)
    at window.Section.initializeTable (section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1:10561)
    at section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1:10489
    at NodeList.forEach (<anonymous>)
    at window.Section.initializeComponents (section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1:10324)
    at new window.Section (section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1:10208)
    at techreport.js?v=2e9f87dae208dd8235b3164b5c8f8067:1:11203
    at NodeList.forEach (<anonymous>)
    at window.TechReport.initializeReport (techreport.js?v=2e9f87dae208dd8235b3164b5c8f8067:1:11182)
    at window.TechReport.initializePage (techreport.js?v=2e9f87dae208dd8235b3164b5c8f8067:1:10547)
i @ section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1
initializeTable @ section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1
(anonymous) @ section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1
initializeComponents @ section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1
window.Section @ section.js?v=b2a3fca9592b64aecf7569ae40b9a80e:1
(anonymous) @ techreport.js?v=2e9f87dae208dd8235b3164b5c8f8067:1
initializeReport @ techreport.js?v=2e9f87dae208dd8235b3164b5c8f8067:1
initializePage @ techreport.js?v=2e9f87dae208dd8235b3164b5c8f8067:1
window.TechReport @ techreport.js?v=2e9f87dae208dd8235b3164b5c8f8067:1
(anonymous) @ comparison:1472
```

src/js/components/filters.js
```diff
@@ -164,14 +159,14 @@ class Filters {
         all.innerHTML = 'ALL';
```
tunetheweb, line 159: Nit, and realise you didn't change this here, but do we ever use HTML here? If not could this be:

```suggestion
        all.textContent = 'ALL';
```

Better for performance and security.

src/js/components/filters.js
```diff
@@ -164,14 +159,14 @@ class Filters {
...
         sortedCategories.forEach((category) => {
           const option = document.createElement('option');
-          if(category.category === select.getAttribute('data-selected')) {
+          if(category === select.getAttribute('data-selected')) {
             option.selected = true;
           }
-          option.value = category.category;
-          option.innerHTML = category.category;
+          option.value = category;
+          option.innerHTML = category;
```
tunetheweb, line 169:
```suggestion
          option.textContent = category;
```

src/js/techreport/tableLinked.js
```diff
@@ -175,7 +151,166 @@ class TableLinked {
...
+    checkbox.setAttribute('data-name', `table-${this.id}`);
+    checkbox.setAttribute('id', `${app}-table-${this.id}`);
+    checkbox.setAttribute('name', `${app}-table-${this.id}`);
+    checkbox.checked = isChecked;
+    checkbox.addEventListener('change', (e) => this.changeCheckbox(e));
+
+    cell.append(checkbox);
+
+    const label = document.createElement('label');
+    label.innerHTML = `Select ${formattedApp}`;
```
tunetheweb, line 180:
```suggestion
    label.textContent = `Select ${formattedApp}`;
```

src/js/techreport/utils/data.js
```diff
@@ -100,6 +102,101 @@ const getLighthouseScoreCategories = (score, brackets) => {
...
+  fetch(url)
+    .then(result => result.json())
+    .then(result => {
+      const category = result[0];
+      const firstTechNr = (pageNr - 1) * rows;
+      const lastTechNr = pageNr * rows;
+      const paginatedTechs = category?.technologies?.slice(firstTechNr, lastTechNr);
+
+      const technologyFormatted = paginatedTechs?.join('%2C')
+        .replaceAll(" ", "%20");
```
tunetheweb, line 140: Should we use `encodeURI` to handle other potential issue characters?

src/js/techreport/utils/data.js
```diff
@@ -100,6 +102,101 @@ const getLighthouseScoreCategories = (score, brackets) => {
...
+
+      const compare = document.querySelector('[data-name="selected-apps"]');
+      compare.setAttribute('href', `/reports/techreport/tech?tech=${technologyFormatted}`);
+
+      const geo = filters.geo.replaceAll(" ", "%20");
+      const rank = filters.rank.replaceAll(" ", "%20");
+      const geoFormatted = geo.replaceAll(" ", "%20");
+      const rankFormatted = rank.replaceAll(" ", "%20");
```
tunetheweb, line 148: Ditto

src/js/techreport/utils/data.js
```diff
@@ -100,6 +102,101 @@ const getLighthouseScoreCategories = (score, brackets) => {
...
+            origins: category.origins,
+            technologies: category?.technologies?.length,
+          },
+        };
+
+        /* Update the pagination info */
+        const current = document.querySelectorAll('[data-page="current"]');
+        const total = document.querySelectorAll('[data-page="total"]');
+        current.forEach(c => c.innerHTML = pageNr);
+        total.forEach(t => t.innerHTML = Math.ceil(category?.technologies?.length / rows));
```
tunetheweb, line 192:
```suggestion
        current.forEach(c => c.textContent = pageNr);
        total.forEach(t => t.textContent = Math.ceil(category?.technologies?.length / rows));
```

I seem obsessed now :-)

tunetheweb (commented):
LGTM once we get CI sorted.

A few minor nits.

tunetheweb (approved):

PR #1013, opened 2025-02-23
added har.fyi link on homepage

max-ostapenko:


max-ostapenko (commented):
Updated the BQ dataset link on the homepage and removed old markdown guide.

And some python linting.

tunetheweb (approved):
LGTM

Issue #1009, opened 2025-02-14
Tech Report: Comparison mixes field and lab performance data


tunetheweb:
From Slack:

> is the comparison view - https://httparchive.org/reports/techreport/tech?tech=ALL,WordPress,Wix,Next.js - using Lighthouse scores for the Perf. and A11Y columns?    Kinda throws me off seeing RUM and synthetic data shown side by side like that -- I suspect people less familiar with web perf won't realize the critical nature of the difference between the two

<img width="1165" alt="Image" src="https://github.com/user-attachments/assets/2c59d813-bc32-445d-85d0-4be2bc7d3d66" />

I think that's fair feedback. The Core Web Vitals are better performance metrics of REAL user data so while it's OK to make the LH Perf score available in the Lighthouse section, I don't think it needs the prominent placement at the top.

Let's drop the LH perf score, make A11Y the first column, and add SEO as the second column.

sarahfossheim:
PR for this opened: https://github.com/HTTPArchive/httparchive.org/pull/1015

PR #1015, opened 2025-02-24
Update performance with seo in comparison table

sarahfossheim:
Fixes: https://github.com/HTTPArchive/httparchive.org/issues/1009

In multi tech comparison table:
- Performance removed from the list
- Accessibility first lighthouse column
- SEO second lighthouse column


tunetheweb (approved):
LGTM!

PR #1019, opened 2025-03-01
Super linter update

max-ostapenko:


tunetheweb (changes requested):
If the PR title is "Bump super-linter/super-linter" (which is what dependabot uses) then it runs the linter on the whole code base to ensure any upgrade works.

I ran it manually and looks like some issues (e.g. CSS upgrade removed a lot of rules so need to remove those rules from our config in this update).

tunetheweb (approved):
LGTM. Thanks!

Issue #1001, opened 2025-01-22
Tech Report: Drop down click area


tunetheweb:
The drop down area interaction is a little strange and unintuitive:

<img width="437" alt="Image" src="https://github.com/user-attachments/assets/ea1ee560-1f4a-43a4-9967-8016c4f07741" />

I would expect the down arrow on the right to be clickable but it's not - you have to click the text field to the left.

Additionally the hover effect of the down arrow is odd. It doesn't go black with you hover of it, but does when you hover the text field to the left and, weirdly, when you hover above it (but then it's still not clickable).

See this video:
https://github.com/user-attachments/assets/d730a35f-ca57-426f-82f9-b3723c81e0e0

rviscomi:
@sarahfossheim will this be obsoleted by https://github.com/HTTPArchive/httparchive.org/issues/993?

sarahfossheim:
@rviscomi @tunetheweb Yes, will be updated along with the other dropdown changes :)

sarahfossheim:
@tunetheweb addressed here: https://github.com/HTTPArchive/httparchive.org/pull/1024

PR #1024, opened 2025-04-01
UI Improvements: dark mode fixes, clearer hover styles, disable dropdowns, and responsive design table

sarahfossheim:
Includes:

- Disable dropdowns until the filters are done loading
- Improvements to the hover effects on the dropdowns: https://github.com/HTTPArchive/httparchive.org/issues/1001
- Icons don't move down to a new line when rescaling the screen anymore
- Improvements to the dark mode (incl. color contrast issues)
- Style selected items in the table (yellow background)
- Style hovers in the tables (blue background)

### DIFF
--- a/src/js/components/drilldownHeader.js
+++ b/src/js/components/drilldownHeader.js
@@ -35,6 +35,9 @@ function setCategories(categories) {
       cellTemplate.textContent = `+ ${more} more`;
       list.appendChild(cellTemplate);
     }
+  } else {
+    const list = document.querySelector('.intro .category-list');
+    list.remove();
   }
 }
 
@@ -42,6 +45,9 @@ function setDescription(description) {
   if(description && description !== "") {
     const descr = document.querySelector('p.app-description');
     descr.textContent = description;
+  } else {
+    const descr = document.querySelector('p.app-description');
+    descr.remove();
   }
 }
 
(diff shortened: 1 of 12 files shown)


tunetheweb:
I'm having trouble seeing any different with this change?:

> Disable dropdowns until the filters are done loading

Can you explain exactly what changed here?

The rest looks good!

sarahfossheim:
@tunetheweb when the page loads the dropdowns for technology/category, geo, and rank are disabled (and with subtle less contrast grayed out styling) until the API returns the list of technologies, geos and ranks to populate it with. Then the disabled attribute is removed and the `<select>` is populated with the `<option>`s.

---

Previously while API is loading:
Dropdown with only one option inside

Currently while API is loading:
List of disabled dropdowns 

After the API call is finished:
list of dropdowns & dropdown opened with many options inside

sarahfossheim:
@tunetheweb that should be updated and ready for another review :)

tunetheweb:
This works well now in the Technologies page, but not the Comparisons page. Should we merge this and fix that in another PR? Or fix it in this PR?

tunetheweb:
SGTM. Yeah noticed the duplicated myself, which is why I didn't do a suggestion :-)

tunetheweb (approved):

PR #1025, opened 2025-04-02
Bugfix: Add the correct techs to the compare button when loading a page

sarahfossheim:
Previous behavior in the categories page: on pageload, regardless if any technologies were selected from the URL or not, the "compare technologies" button above the table compared _all_ technologies on the page. Additionally, the URL it led to was formatted wrong (one long string with `%2C` as a separator instead of commas), and as a result the compare page didn't work.

Updated code: 
- Checks first if techs are selected or not
- If any techs are selected, those are used
- If not, all techs on the page are used
- Then the list is separated by comma before being encoded

Result:
- Enter the page with no techs selected → Link leads to: Comparison page with all techs in the table
- Enter the page with one or more techs selected → Link leads to: Comparison page with the selected techs
- The actual techs are loaded in the comparison page

### DIFF
--- a/src/js/components/filters.js
+++ b/src/js/components/filters.js
@@ -73,7 +73,8 @@ class Filters {
     // url.hash = '#report-content';
 
     /* Update the url */
-    location.href = url;
+    const styledUrl = url.href.replaceAll('%2C', ',');
+    location.href = styledUrl;
   }
 
   /* Update the list of technologies */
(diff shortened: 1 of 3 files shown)


sarahfossheim:
yes, will fix that, looks nicer when it's consistent!

tunetheweb (approved):
LGTM thanks!

Issue #1036, opened 2025-04-16
Tech Report: Bug in selecting additional technologies


tunetheweb:
Looks to be related to disabling drop downs in #1024 but now can't add additional categories.

Steps to reproduce:
1. Go to https://httparchive.org/reports/techreport/tech?geo=ALL&rank=ALL&tech=ALL
2. Add another technology

<img width="326" alt="Image" src="https://github.com/user-attachments/assets/934d91aa-88b7-422d-afb2-1ecc540385be" />

As can be seen the new ones is added as disabled by default with the first 1C-Bitrix tecnology and no way to change it :-(

PR #1037, opened 2025-04-16
Bugfix: Remove disabled state when adding new tech

sarahfossheim:
Last PR introduced a bug: when clicking the "add new tech" button, a disabled dropdown was added to the filters. 
Updated: the `disabled` attribute is removed from new dropdowns added.

Fixes: https://github.com/HTTPArchive/httparchive.org/issues/1036


tunetheweb (approved):
LGTM - thanks for the quick fix!

PR #1043, opened 2025-04-30
Use custom multi select dropdowns for the tech selector

sarahfossheim:


sarahfossheim:
from meeting:
- move icon to right in button

src/js/techreport/combobox.js
```diff
@@ -0,0 +1,249 @@
...
+    this.focusedOptionStr = '';
+    this.selected = [];
+    this.maxSelected = 10;
+
+    this.keyCodes = {
+      keyUp: 38,
+      keyDown: 40,
+      ESC: 27,
+      enter: 13,
+    };
```
tunetheweb, line 17: I'm not sure we need this? More explanation below.

src/js/techreport/combobox.js
```diff
@@ -0,0 +1,249 @@
...
+          const option = options[this.focusedOption];
+          const selected = option.getAttribute('aria-selected');
+          if(selected === 'true') {
+            this.unselectElement(option.dataset.name);
+          } else {
+            this.selectElement(option);
+          }
+        }
+    }
+  }
```
tunetheweb, line 149: Is there a reason we can't use this?

```suggestion
    const key = e.key;
    switch(key) {
      case 'ArrowDown':
        const nextKey = this.focusedOption + 1;
        if(nextKey < options.length) {
          this.focusedOption = nextKey;
          this.updateHighlightedOption();
        }
        break;
      case 'ArrowUp':
        const previousKey = this.focusedOption - 1;
        if(previousKey > -1) {
          this.focusedOption = previousKey;
          this.updateHighlightedOption();
        }
        break;
      case 'Escape':
        this.hideOptions();
        break;
      case 'Enter':
        e.preventDefault();
        const listbox = this.element.querySelector('[role="listbox"]');
        if(!listbox.classList.contains('hidden')) {
          const option = options[this.focusedOption];
          const selected = option.getAttribute('aria-selected');
          if(selected === 'true') {
            this.unselectElement(option.dataset.name);
          } else {
            this.selectElement(option);
          }
        }
    }
  }
```

tunetheweb (commented):
This looks great! Will give it a full review when you're finished but for now spotted one this.

src/js/techreport/combobox.js
sarahfossheim: updated 👍🏻

src/js/techreport/combobox.js
```diff
@@ -0,0 +1,261 @@
...
+      option.setAttribute('role', 'option');
+      option.setAttribute('aria-selected', 'false');
+      option.dataset.name = row;
+      option.dataset.key = index;
+      option.textContent = row;
+      option.id = `${this.element.dataset.id}-${row}`;
+      const logo = document.createElement('img');
+      logo.setAttribute('alt', '');
+      logo.setAttribute('src', `https://cdn.httparchive.org/static/icons/${row}.png`);
+      logo.setAttribute('height', '14px');
```
tunetheweb, line 34: Height attributes are not in pixels:

```suggestion
      logo.setAttribute('height', '14');
```

This is causing the Lighthouse fail

src/js/techreport/combobox.js
```diff
@@ -0,0 +1,261 @@
...
+      logo.setAttribute('width', 'auto');
```
tunetheweb, line 35: Let's add a `loading=lazy` attribute too so we don't eagerly download ~3700 logos

src/js/techreport/combobox.js
```diff
@@ -0,0 +1,261 @@
...
+    // Add options in the dropdown list
+    const listbox = this.element.querySelector('[role="listbox"]');
+    rows.forEach((row, index) => {
+      const option = document.createElement('div');
+      option.setAttribute('role', 'option');
+      option.setAttribute('aria-selected', 'false');
+      option.dataset.name = row;
+      option.dataset.key = index;
+      option.textContent = row;
+      option.id = `${this.element.dataset.id}-${row}`;
```
tunetheweb, line 30: Need to strip out spaces. And maybe other characters?

```suggestion
      option.id = `${this.element.dataset.id}-${row.replaceAll(' ','-')}`;
```

src/js/techreport/combobox.js
```diff
@@ -0,0 +1,261 @@
...
+      const logo = document.createElement('img');
+      logo.setAttribute('alt', '');
+      logo.setAttribute('src', `https://cdn.httparchive.org/static/icons/${row}.png`);
```
tunetheweb, line 33: This assumes the icon is the technology name, but that's not always the case.

For example:
- "Yandex SmartCaptcha" - "yandex_smartcaptcha.png"

It also doesn't work when a technology reuses an icon. For example:

```
  "Yandex.Cloud": {
    "cats": [
      62
    ],
    "description": "Yandex.Cloud is a public cloud platform where companies can create and develop projects using Yandex's scalable computing power, advanced technologies, and infrastructure.",
    "icon": "Yandex.Cloud.svg",
    "pricing": [
      "payg"
    ],
    "saas": false,
    "website": "https://cloud.yandex.com/en/"
  },
  "Yandex.Cloud CDN": {
    "cats": [
      31
    ],
    "description": "Yandex.Cloud CDN helps you streamline static content delivery for your web service.",
    "dom": "[href*='storage.yandexcloud.net'], [src*='storage.yandexcloud.net']",
    "icon": "Yandex.Cloud.svg",
    "implies": "Yandex.Cloud",
    "website": "https://cloud.yandex.com/en/services/cdn"
  },
```

Can we pull in the actual icon name from the API?

tunetheweb (changes requested):

src/js/techreport/combobox.js
sarahfossheim: Yes, but then we need to drop the `?onlyname` from the request to the technologies endpoint, and we added it a while back to speed up the API response 😅

So options are:
* Fetch the icons for the dropdowns using `/v1/technologies`, have a slower API response
* Keep the icons in the dropdowns based on the names (`/v1/technologies?onlyname`), have some icons not show up
* Remove the icon from the dropdowns

@max-ostapenko it seems that your open PR https://github.com/HTTPArchive/httparchive.org/pull/1048 will speed up the technologies endpoint, do you think it will be ok to use it again without the `onlyname` for the dropdowns?

Issue #1035, opened 2025-04-16
Tech Report: Add heading links


tunetheweb:
It would be nice to be able to do this:

https://httparchive.org/reports/techreport/tech?tech=WordPress&geo=ALL&rank=ALL#adoption

Also we should have anchor links. In the Web Almanac we show those on hover so maybe could be consistent with that style?:

https://almanac.httparchive.org/en/2024/cdn#what-is-a-cdn

<img width="257" alt="Image" src="https://github.com/user-attachments/assets/7cce1de6-a2b9-44c4-87df-7ef86d6503ed" />

We do similar in reports, but without the symbol:
https://httparchive.org/reports/state-of-the-web#reqTotal

<img width="527" alt="Image" src="https://github.com/user-attachments/assets/e32d743b-80db-48ce-ab1c-e28f5106a1e1" />

tunetheweb:
Fixed in #1044

PR #1044, opened 2025-04-30
Turn headings into links

sarahfossheim:
* Headings are now links, linking to the section
* Hover/focus over the heading shows the anchor symbol

As requested here: https://github.com/HTTPArchive/httparchive.org/issues/1035

### DIFF
--- a/static/css/techreport/techreport.css
+++ b/static/css/techreport/techreport.css
@@ -229,17 +229,6 @@ nav {
   color: inherit;
 }
 
-/* .anchor:is(:hover, :focus-visible) {
-  position: relative;
-}
-
-.anchor:is(:hover, :focus-visible)::after {
-  content: "🔗" / "";
-  display: inline-block;
-  margin-left: 0.25em;
-  font-size: 1em;
-} */
-
 /* Filter info */
 .meta {
   font-size: 0.75rem;
--- a/static/css/techreport/techreport.css
+++ b/static/css/techreport/techreport.css
@@ -229,18 +229,16 @@ nav {
   color: inherit;
 }
 
-.anchor:is(:hover, :focus-visible) {
+/* .anchor:is(:hover, :focus-visible) {
   position: relative;
 }
 
 .anchor:is(:hover, :focus-visible)::after {
-  content: "🔗";
-  display: block;
-  position: absolute;
-  right: -1.3em;
-  top: 0.4em;
-  font-size: 1.5rem;
-}
+  content: "🔗" / "";
+  display: inline-block;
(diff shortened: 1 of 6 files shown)


static/css/techreport/techreport.css
```diff
@@ -224,6 +224,24 @@ nav {
...
+.anchor {
+  color: inherit;
+}
+
+.anchor:is(:hover, :focus-visible) {
+  position: relative;
+}
+
+.anchor:is(:hover, :focus-visible)::after {
+  content: "🔗";
```
tunetheweb, line 237: Remove it from the accessible name (with fallback as only supported in iOS and Firefox last year - I'm OK to drop the fallback too WDYT?):

```suggestion
  content: "🔗";
  content: "🔗" / "";
```

static/css/techreport/techreport.css
```diff
@@ -224,6 +224,24 @@ nav {
...
+  display: block;
+  position: absolute;
+  right: -1.3em;
+  top: 0.4em;
```
tunetheweb, line 241: Would `display: inline-block` be better as then don't need to position it absolutely?

tunetheweb (changes requested):
Thanks for this.

Some observations:
1. There's an annoying shift when you hover the link as the underling is drawn. Can we avoid that?
2. I think we should do H3's too? Was a little surprised only some headings where clickable/linkable until I realised that was the reason.
3. Should do "Feedback?" and "Accessibility" h2 headings too.

static/css/techreport/techreport.css
sarahfossheim: I actually ended up removing the anchor symbol all together to make it consistent with the other reports, which don't have them

sarahfossheim:
@tunetheweb 

> There's an annoying shift when you hover the link as the underling is drawn. Can we avoid that?

You mean a shift in layout? Which browser/OS? I don't have it on Chrome/Safari on Mac 😅

> I think we should do H3's too? Was a little surprised only some headings were clickable/linkable until I realised that was the reason.

H3s are now clickable too :)

> Should do "Feedback?" and "Accessibility" h2 headings too.

Also done :)

tunetheweb:
> > There's an annoying shift when you hover the link as the underling is drawn. Can we avoid that?
> 
> You mean a shift in layout? Which browser/OS? I don't have it on Chrome/Safari on Mac 😅

Hmmm can't repeat this now! Maybe I didn't refresh my cache.

tunetheweb (approved):
LGTM!

PR #1048, opened 2025-05-11
New Tech report API backend

max-ostapenko:
A new version of API backend deployed to DEV.
I had much better IO with nodeJS for BQ->Firestore import, so decided to test here too.

@sarahfossheim would be great if you could run this branch to check if I missed anything.
If all good, then I'll deploy to PROD and update the `baseApi`.

Addressed limitations of the current implementation:
- express js requires much less CPU for simple data passing
- with a single backend we need to keep just this one ON, always ready to serve (also simpler source code for the backend and infra)

I deployed it in `us-central1` as it has 'eco' label. I'm not aware of any advantages of `us-east` (I can move it easily if there are any).
Region picker also recommends central, even if we disregard the carbon footprint.
image

Some lab tests for the biggest response:
Screenshot 2025-05-11 at 21 34 21

I'm not sure it's possible to get this under 1s ;)

Note: existing DB is in `us-east`, the new DB instance is also in `us-central1`.


max-ostapenko:
@tunetheweb please let me know your idea about this.
If OK I'll be able to migrate `/technologies` and `/categories` (the slowest) to the new DB already.

tunetheweb:
I'm easy on the technology. Node is probably better understood than Python but either works.

Just to understand the output from the endpoints is exactly the same? It's all just the behind the scenes stuff you're changing?

I'm confused about the following line and what exactly it means?:

> with a single backend we need to keep just this one ON, always ready to serve (also simpler source code for the backend and infra)

So are we saying the current (python?) infrastructure fires ups several instances in demand in multiple reasons? And the new one will have a single instance in one region?

So I guess there's pluses and minuses to each:

**Multiple, on-deman services**

- Scales better
- Lag during start up
- If services crash then another service will be started
- Releases might be slower to roll out?

**Single, always-on service**

- Lower latency
- Scalability issues?
- Slower outside of USA?
- What happens when it crashes?
- What happens when doing a release?

Is that a fair summary or are there other concerns?

tunetheweb:
OK seems I misunderstood. I thought the current python endpoints were replicated around the global (i.e. they were edge services). But that's not the case. We have one python server per endpoint (and all are location in one of two regions). But you propose joining them into one server for all end points - all running in one region.

So we're not losing global reach/stablitity because we never had that. One for future maybe!

Anyway, in that case I'm happy with this proposal. But do look at whether Express is adding any value.

max-ostapenko:
@tunetheweb I updated the endpoint to production.

I don't know a way to release without disruption. The website update changes the domain, and the API release moves the API endpoints.
The lowest traffic seems to be at 6-7pm UTC, I can release it then.

Regarding your notes:
> Lower latency

for `categories` and `technologies` in particular

> Scalability issues?

less than with python, requires much less resources

> Slower outside of USA?

based on GCP regions recommender should be faster for top5 countries
[Noted a TODO](https://github.com/HTTPArchive/tech-report-apis/issues/21#issuecomment-2881748456) to check out Global Load Balancer + CDN

> What happens when it crashes?

The impact depends on the issue, but a bug in some core code may bring it to a total failure.
We have incident alerts for this in the Slack channel.

> What happens when doing a release?

Release is much faster as we build and update only one function.

PR #1057, opened 2025-06-02
Lighthouse changes

tunetheweb:
Makes a few changes to Lighthouse in the new tech report
- Changes the Performance SEO and also moves a11y up a notch - similar to #1009
- Consistently uses capital for Lighthouse in labels since it's a brand name
- Consistently uses capital letter for category (Accessibility not accessibility) when talking about the audits categories.
- Fixes some links

### DIFF
--- a/config/techreport.json
+++ b/config/techreport.json
@@ -122,23 +122,23 @@
           },
           {
             "endpoint": "lighthouse",
-            "category": "performance",
+            "category": "accessibility",
             "metric": "median_score_pct",
-            "label": "Lighthouse performance",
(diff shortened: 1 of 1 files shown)


tunetheweb:
I've also turned off these audits temporarily since [they are incompatible with the latest version of Chrome until Lighthouse-ci is upgrade](https://github.com/GoogleChrome/lighthouse-ci/issues/1106):

- "lcp-lazy-loaded": "off",
- "non-composited-animations": "off",
- "prioritize-lcp-image": "off",

FYI, @max-ostapenko this is what you spotted last week.

PR #1064, opened 2025-06-05
Fix Highcharts export

### LINKED ISSUE #1062: Tech Report: Downloading chart results in CSP error
When looking at a [chart](https://httparchive.org/reports/techreport/tech?good-cwv-over-time=CLS&client=desktop&tech=ALL,WordPress,Shopify,Wix,Squarespace,Joomla,Webflow,Drupal&geo=ALL&rank=ALL), I tried to download one of the charts via the menu. But when I did so, nothing happened. I looked at the console and found a CSP error:

> Fetch API cannot load https://export.highcharts.com/. Refused to connect because it violates the document's Content Security Policy.

max-ostapenko:
- [x] fixed Highcharts export, closes #1062 
- [x] migrated to Reports API Prod
- [x] removed Fontawesome, seems to be an outdated host


tunetheweb (approved):
It looks to me like Lighthouse is crashing in CI. Not sure why, but I can't see it being related to your change. We'll assume it's flakey for now and I'll take a closer look later if it's still happening.

FYI, I did add one other typo fix I spotted when investigating.

Good to merge and release this.

tunetheweb:
I think the 404s are due to this: https://github.com/HTTPArchive/httparchive.org/pull/1043#discussion_r2089615072 that we said we'd fix in a follow up PR once your new API changes were live (which they are now I presume!)

Issue #1067, opened 2025-06-08
Tech Report: MoM icons rendered incorrectly


max-ostapenko:
Example: https://httparchive.org/reports/techreport/tech?tech=33Across&geo=ALL&rank=ALL&client=desktop

Switch clients to reproduce.

![Image](https://github.com/user-attachments/assets/725e83f4-db1b-485c-bef2-f6075b64ab68)

sarahfossheim:
Good catch! When updating the styling of the MoM we add class names (good/bad for the color, positive/negative for the direction), but never delete the previous one:

``` 
changeSlot.classList.add(styling.color, styling.direction)
```

On it!

PR #1069, opened 2025-06-08
#1067: Overwrite MoM classnames instead of adding to the list

sarahfossheim:
## Issue
Styling was being added to the month over month components by adding to the classList:
```
changeSlot.classList.add(styling.color, styling.direction)
```
or:
```
changeSlot.classList.add('monthchange', styling.color, styling.direction)
```

**As a result**, the icons were added twice when switching between mobile and desktop (which doesn't reload the page or clear out the existing components).

## Fix

Instead of adding to the `classList`, the `className` now gets overwritten:

```
changeSlot.className = `month change ${styling.color} ${styling.direction}`.

### DIFF
--- a/src/js/techreport/summaryCards.js
+++ b/src/js/techreport/summaryCards.js
@@ -73,7 +73,7 @@ class SummaryCard {
         });
       }
 
-      if(latestChange && latestChange.string && latestChange.perc) {
+      if(latestChange && latestChange.string && latestChange.perc != null) {
         const changeSlot = card.querySelector('[data-slot="change"]');
         const changeMeaning = changeSlot?.dataset?.meaning;
 

### CODE REVIEW

File: src/js/techreport/table.js [hunk 1]
```diff
@@ -194,7 +194,7 @@ function updateTable(id, config, appConfig, apps, data) {
       }
 
       if(column.styling) {
-        wrapper.classList.add('monthchange', column.styling.direction, column.styling.color);
+        wrapper.className = `monthchange ${styling.color} ${styling.direction}`;
```
tunetheweb, line 197: Doesn't this need to be:

```suggestion
        wrapper.className = `monthchange ${column.styling.color} ${column.styling.direction}`;
```

tunetheweb (changes requested):

src/js/techreport/table.js
tunetheweb: Fixed this in the merge with `main` from #1068

PR #1087, opened 2025-07-12
Safe int conversion for query parameters

max-ostapenko:
- [x] safe int conversion for numerical query params

  Lots of 500 in logs: https://cloudlogging.app.goo.gl/tUXGV92PUkmsegiPA

  And one more argument to switch off security scanner.

- [x] reverted some false linting pushed to `main` by mistake.

### DIFF
--- a/server/routes.py
+++ b/server/routes.py
@@ -99,7 +99,7 @@ def techreportlanding(page_id):
     requested_geo = request.args.get("geo") or "ALL"
     requested_rank = request.args.get("rank") or "ALL"
     requested_category = request.args.get("category") or "CMS"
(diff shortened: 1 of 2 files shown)


tunetheweb (approved):
Not totally convinced that ignoring bad params is actually a good thing to do. However I also don't think sending a 500 error is a good thing to do either (400 is probably more accurate).

Anyway, don't feel strongly about it, so if if it makes our life easier then happy to go with this.

max-ostapenko:
Yeah, I'd like to do 400, but 99.9% of our requests are coming from security scanner.
So will mute them for now.

PR #1142, opened 2025-10-15
Move meta data above the chart

### LINKED ISSUE #1127: Tech Report: Consider adding selected Technology, Geography to the graph as text
Currently in the new report it's hard to determine from a screenshot what Geo/Tech filters were applied.
In the old report we added "Geo/Rank" as a small text to the image to clarify this, and the tech filter is right above the graph

In the new report its hard to understand from screenshot what technology/geo/rank was filtered

Image

Image

tunetheweb:
Fixes #1127

### DIFF
--- a/templates/techreport/components/filter_meta.html
+++ b/templates/techreport/components/filter_meta.html
@@ -8,4 +8,10 @@
   <li>
     Rank: <span data-slot="rank">{{ request.args.get('rank', '')  or 'ALL' }}</span>
   </li>
+  {# if only one technology (including ALL), then legend is not shown so add label #}
+  {% if ',' not in request.args.get('tech', '') %}
+  <li>
+    Technologies: <span data-slot="tech">{{ request.args.get('tech', '')  or 'ALL' }}</span>
+  </li>
+  {% endif %}
 </ul>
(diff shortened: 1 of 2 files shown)


templates/techreport/components/filter_meta.html
```diff
@@ -8,4 +8,10 @@
   <li>
     Rank: <span data-slot="rank">{{ request.args.get('rank', '')  or 'ALL' }}</span>
   </li>
+  {# if only one technology (including ALL), then legend is not shown so add label #}
+  {% if ',' not in request.args.get('tech', '') %}
```
alonkochba: I would do this check around Client: as well (but hide label if its single technology)

otherwise you end up with `Client: Mobile`, even though it shows both mobile AND desktop when a single technology is selected

templates/techreport/components/filter_meta.html
```diff
@@ -8,4 +8,10 @@
...
+  <li>
+    Technologies: <span data-slot="tech">{{ request.args.get('tech', '')  or 'ALL' }}</span>
```
alonkochba: `Technologies` -> `Technology:`?
Or did you leave it as technologies for the case of ALL?

tunetheweb: Yeah. And think it's fine to be plural as indicates it is possible to select multiples ones (unlike Geo or Rank)

PR #1241, opened 2026-04-07
Add Core Web Vitals distribution histogram to tech report

### LINKED ISSUE #1147: Tech Report: Histograms per KPI
It's very useful to see histograms per KPI
A lot of very interesting data hides in the percentiles movement, for example green sites becoming much greener. 

Simple single month (ignore label):

Image

Monthly comparisons (advanced):

Image

alonkochba:
## Summary

Adds two collapsible panels below the Core Web Vitals timeseries chart:

- **Histogram** — distribution of origins across performance buckets for the selected metric, with good/needs-improvement/poor color zones, threshold plotlines, and an overflow bucket for the long tail
- **Geographic breakdown** — existing section, moved here as a toggle

Both panels share a single button bar (alongside `Show table`) with mutual exclusion — opening one closes the other. Content is lazy-loaded on first open.

The metric selector from the pass-rate timeseries now drives all three components (timeseries, geo breakdown, histogram) via a shared `cwv-metric-change` event and the `good-cwv-over-time` URL param. Deep-linking with `?good-cwv-over-time=CLS` applies to all three. Titles update dynamically (`LCP histogram`, `INP geographic breakdown`, etc.). The histogram resolves `overall` to `LCP` since it shows one metric at a time.

Also scopes the global `.highcharts-point` CSS rule to line/spline series only, so column chart bar colors render correctly.

## Depends on

- HTTPArchive/tech-report-apis#105 — adds the `/v1/cwv-distribution` BigQuery endpoint

Closes #1147

### DIFF
--- a/config/techreport.json
+++ b/config/techreport.json
@@ -734,6 +734,18 @@
             { "label": "Good FCP", "value": "FCP" },
             { "label": "Good TTFB", "value": "TTFB" }
           ]
+        },
+        "cwv_distribution": {
+          "id": "cwv_distribution",
+          "title": "Core Web Vitals distribution",
+          "description": "How origins are distributed across performance buckets for individual Core Web Vitals metrics. Green, orange, and red zones indicate good, needs improvement, and poor thresholds respectively.",
+          "metric_options": [
+            { "label": "LCP", "value": "LCP" },
+            { "label": "CLS", "value": "CLS" },
+            { "label": "INP", "value": "INP" },
+            { "label": "FCP", "value": "FCP" },
+            { "label": "TTFB", "value": "TTFB" }
+          ]
         }
       }
     },
--- a/config/techreport.json
+++ b/config/techreport.json
@@ -737,8 +737,8 @@
         },
         "cwv_distribution": {
           "id": "cwv_distribution",
-          "title": "Core Web Vitals distribution",
-          "description": "How origins are distributed across performance buckets for individual Core Web Vitals metrics. Green, orange, and red zones indicate good, needs improvement, and poor thresholds respectively.",
+          "title": "Core Web Vitals histograms",
+          "description": "How origins are distributed for individual Core Web Vitals metrics. Green, orange, and red zones indicate good, needs improvement, and poor thresholds respectively.",
           "metric_options": [
             { "label": "LCP", "value": "LCP" },
             { "label": "CLS", "value": "CLS" },
--- a/config/techreport.json
+++ b/config/techreport.json
@@ -15,6 +15,13 @@
       "INP",
       "TTFB"
     ],
+    "cwv_thresholds": {
+      "LCP":  [{ "value": 2500, "label": "Good" }, { "value": 4000, "label": "Needs improvement" }],
+      "FCP":  [{ "value": 1800, "label": "Good" }, { "value": 3000, "label": "Needs improvement" }],
+      "TTFB": [{ "value": 800,  "label": "Good" }, { "value": 1800, "label": "Needs improvement" }],
+      "INP":  [{ "value": 200,  "label": "Good" }, { "value": 500,  "label": "Needs improvement" }],
+      "CLS":  [{ "value": 0.1,  "label": "Good" }, { "value": 0.25, "label": "Needs improvement" }]
+    },
     "lighthouse_subcategories": [
       "accessibility",
       "best_practices",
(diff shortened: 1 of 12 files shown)


config/techreport.json
```diff
@@ -734,6 +734,18 @@
             { "label": "Good FCP", "value": "FCP" },
             { "label": "Good TTFB", "value": "TTFB" }
           ]
+        },
+        "cwv_distribution": {
+          "id": "cwv_distribution",
+          "title": "Core Web Vitals distribution",
+          "description": "How origins are distributed across performance buckets for individual Core Web Vitals metrics. Green, orange, and red zones indicate good, needs improvement, and poor thresholds respectively.",
```
tunetheweb: We use histograms on the main site. ([example](https://httparchive.org/reports/state-of-the-web?start=latest&view=list))


```suggestion
          "id": "cwv_distribution",
          "title": "Core Web Vitals histograms",
          "description": "How origins are distributed for individual Core Web Vitals metrics. Green, orange, and red zones indicate good, needs improvement, and poor thresholds respectively.",
```

templates/techreport/components/cwv_distribution.html
```diff
@@ -0,0 +1,44 @@
...
+    <div class="component-filters">
+      <div class="subcategory-selector-wrapper">
+        <div class="position-wrapper">
+          <label for="cwv-distribution-metric-selector">Metric</label>
+          <select
+            class="cwv-distribution-metric-selector"
+            id="cwv-distribution-metric-selector"
+          >
+            {% for option in cwv_distribution_config.metric_options %}
+              {% if option.value == 'LCP' %}
```
tunetheweb: I don't understand why this is restricted to LCP?

templates/techreport/drilldown.html
```diff
@@ -67,13 +67,46 @@ <h3>{{ tech_report_page.config.good_cwv_summary.title }}</h3>
...
-          {% include "techreport/components/geo_breakdown.html" %}
+          {% set ts_id = tech_report_page.config.good_cwv_timeseries.id %}
+          <div class="cwv-button-bar">
+            <button class="btn show-table cwv-show-table-btn" type="button" data-id="{{ ts_id }}">Show table</button>
+            {% if tech_report_page.config.geo_breakdown %}
+              {% include "techreport/components/geo_breakdown.html" %}
+            {% endif %}
+            {% if tech_report_page.config.cwv_distribution %}
+              {% include "techreport/components/cwv_distribution.html" %}
+            {% endif %}
```
tunetheweb, line 79: The button order becomes funny here because you have:

- Show table button
- Show Geo button
  - Geo data
- Show Dist button
  - Dist data

It should really be:

- Show table button
- Show Geo button
- Show Dist button
  - Geo data or Dist data

So you can still tab between all the buttons.

static/css/techreport/techreport.css
```diff
@@ -2336,3 +2340,91 @@ path.highcharts-tick {
...
+  flex-basis: 100%;
+  order: 1;
+}
+
+/* CWV Distribution histogram */
+.cwv-distribution-wrapper {
+  margin-top: 1rem;
+  border-top: 1px solid var(--color-border);
+  padding-top: 1rem;
+}
```
tunetheweb, line 2364: Not sure where in the CSS but could do with fixing the selector padding on smaller screens:

Image

src/js/techreport/cwvDistribution.js
```diff
@@ -0,0 +1,285 @@
...
+  dark:  { good: '#0CCE6B', needsImprovement: '#FBBC04', poor: '#FF6659', text: '#ccc', gridLine: '#444' },
+};
+
+class CwvDistribution {
+  // eslint-disable-next-line no-unused-vars -- pageConfig, config, data satisfy the Section component contract
+  constructor(id, pageConfig, config, filters, data) {
+    this.id = id;
+    this.pageFilters = filters;
+    this.distributionData = null;
+    this.selectedMetric = 'LCP';
```
tunetheweb: I think it would be good to link this chart to the pass rate chart above. So if you change one selector, it changes the other too.

That would also allow deeplinking to the CLS distribution for example (whcih you can't do now as it always reverts to LCP).

Alternatively you could have them independent and two URL params, but think having them depdent on each other makes sense.

tunetheweb (changes requested):

src/js/techreport/cwvDistribution.js
alonkochba: Sure, I could do that. What should we display when "Good Core Web Vitals" is selected though? LCP?
(histograms is single metric only)

tunetheweb: Yeah LCP (as you alsready implemented is fine).

tunetheweb (approved):

PR #1264, opened 2026-06-05
Implement lazy loading for combobox technology icons using IntersectionObserver

max-ostapenko:
The log entries indicated that an incoming request to the report-api backend service was denied with a 403 Forbidden status code by a Google Cloud Armor security policy. Specifically, the request from particular IPs exceeded the rate limit threshold defined in the policy named report-api-backend-security-policy.

This means:
- The individual client IPs make an abnormally high volume of requests.
- The default rate limit threshold configured in the Cloud Armor policy is too restrictive for our users.

The solution:
Optimized the technology dropdown list to prevent fetching 2,000+ technology icons simultaneously on load or filter, avoiding overwhelming both the client browser's connection limit and the CDN.

### DIFF
--- a/src/js/techreport/combobox.js
+++ b/src/js/techreport/combobox.js
@@ -21,6 +21,8 @@ class ComboBox {
     // Add options in the dropdown list
     const listbox = this.element.querySelector('[role="listbox"]');
 
+    // Typing in the filter box does work for loading=lazy
+    // so use a manual IntersectionObserver for now.
     const hasObserver = typeof IntersectionObserver !== 'undefined';
     if (hasObserver) {
       if (!this.observer) {
--- a/src/js/techreport/combobox.js
+++ b/src/js/techreport/combobox.js
@@ -20,6 +20,33 @@ class ComboBox {
     const rows = data || this.data;
     // Add options in the dropdown list
     const listbox = this.element.querySelector('[role="listbox"]');
+
+    const hasObserver = typeof IntersectionObserver !== 'undefined';
+    if (hasObserver) {
+      if (!this.observer) {
+        this.observer = new IntersectionObserver((entries) => {
+          entries.forEach(entry => {
+            if (entry.isIntersecting) {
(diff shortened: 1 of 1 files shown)


src/js/techreport/combobox.js
```diff
@@ -20,6 +20,33 @@ class ComboBox {
     const rows = data || this.data;
     // Add options in the dropdown list
     const listbox = this.element.querySelector('[role="listbox"]');
+
+    const hasObserver = typeof IntersectionObserver !== 'undefined';
```
tunetheweb, line 26: This `loading=lazy` attribute means all this code shouldn't be needed.

However, it seems to me Safari is not honouring this and loads everything 😔 It seems to work fine in Chrome and Firefox. But then we see the error in Lighthouse too, which uses Chrome 🤔

But this does seem to fix it for Safari. Is that why you're adding this? Or are there other reasons? I think we should add a comment here.

```suggestion
    // Safari doesn't seem to honour `loading=lazy and loads all icons
    // so use a manual IntersectionObserver for now.
    const hasObserver = typeof IntersectionObserver !== 'undefined';
```

max-ostapenko: It doesn't work for me in Chrome. Try filtering in the search bar inside the dropdown.
Seems lazy loading doesn't cover interactive use-cases.

I didn't test other browsers TBH.

max-ostapenko: I think every time somebody tries to filter for the technologies the browsers and CDN get overwhelmed.

tunetheweb: Ohhh interesting. I didn't test that.

OK let's change the comment and then merge this.

```suggestion
    // Typing in the filter box does work for loading=lazy
    // so use a manual IntersectionObserver for now.
    const hasObserver = typeof IntersectionObserver !== 'undefined';
```

tunetheweb (approved):

PR #1278, opened 2026-06-29
Revert "Bump highcharts from 11.4.8 to 13.0.0"

tunetheweb:
Reverts HTTPArchive/httparchive.org#1266


tunetheweb:
FYI @max-ostapenko . We don't upgrade highcharts.

max-ostapenko:
Yeah, now after studying https://www.highcharts.com/license-eula I see that we need to explicitly apply for a license.

Did you try to reach out?

PR #1299, opened 2026-07-31
Add delay to test reports

tunetheweb:
To avoid hitting rate limits


max-ostapenko (approved):
I'm not sure it will satisfy the rate limits, but let's check

tunetheweb:
Did a few test runs before asking for approval and all working now.

PR #1328, opened 2026-09-13
Lazy load CWV distribution histogram on user click

max-ostapenko:
## Summary

The `/v1/cwv-distribution` endpoint queries BigQuery directly on the backend and is costly to execute on every page load. Because the histogram chart in the tech report is collapsed by default, this PR optimizes data loading to only fetch when requested by the user.

## Changes

- **Lazy load on button click**:
  - Prevent initial fetch of CWV distribution data on page load.
  - Fetch distribution data only when the user expands the section via the "Show histogram" button (`#cwv-distribution-btn`).
  - Cache fetched data so subsequent collapse/expand toggles do not re-fetch.
- **Maintain filter update synchronization**:
  - When the histogram is open and the user clicks the "Update" button (`#submit-form`), append `#section-cwv_distribution` to the URL so the section remains open after reload.
  - Automatically fetch fresh distribution data matching the newly applied filters upon reload.
  - If the histogram is collapsed, filter updates do not trigger any distribution fetch.
- **Race condition guards**:
  - Added `isLoading` state and URL tracking (`getUrl()`) to prevent duplicate in-flight requests between date updates and toggle events.

## Verification

Verified in browser via dev server (`http://localhost:8080/reports/techreport/tech?tech=WordPress`):
1. Initial page load: 0 requests to `/v1/cwv-distribution`.
2. Expand click: Exactly 1 request sent, chart rendered with Core Web Vitals threshold zones.
3. Collapse & re-open: Reused cached data, 0 additional requests.
4. Filter update while visible: Preserved section state and reloaded data for updated filter parameters.
5. Filter update while hidden: Remained hidden with 0 requests sent.


src/js/techreport/cwvDistribution.js
```diff
@@ -83,6 +87,17 @@ class CwvDistribution {
...
+    const submitBtn = document.getElementById('submit-form');
+    if (submitBtn) {
+      submitBtn.addEventListener('click', () => {
+        if (this.root && !this.root.classList.contains('hidden')) {
+          const url = new URL(window.location.href);
+          url.hash = `#section-${this.id}`;
+          window.history.replaceState(null, null, url);
+        }
+      }, true);
+    }
```
tunetheweb: Confused why adding lazy-loading to one chart requires this addition?

tunetheweb (approved):

src/js/techreport/cwvDistribution.js
max-ostapenko: The form rebuilds the URL using filter settings - I was trying to keep the hash.
You're right - it works even better without this piece, as hash is persisted.

Should I replace
[// /* Scroll to the report content */](https://github.com/HTTPArchive/httparchive.org/blob/ec6bdf013bb82408195940aa97b6246b13ee04b5/src/js/components/filters.js#L94)
with a note about persisted hash?

tunetheweb: SGTM

PR #1332, opened 2026-09-14
Merge har.fyi under docs

max-ostapenko:
This PR merges `har.fyi` content directly into `httparchive.org` under the `/docs/` path:
- Adds Astro Starlight integration (`@astrojs/starlight`, `starlight-links-validator`, `starlight-llms-txt`)
- Migrates all guides, reference docs (tables, blobs, structs, custom metrics, functions), and assets into `src/content/docs/`
- Integrates custom Starlight header matching HTTP Archive branding and search
- Updates sitemap and timestamp generation to index doc pages
- Updates site navigation and references pointing to `har.fyi` to relative `/docs/` paths


tunetheweb (changes requested):
This is looking much better on desktop but still have the double hamburger menu issues on mobile:

Image

tunetheweb:
Looking better and better. It definitely feels like it's all part of one site now with the same menu persistent across all pages. It's not quite one case, but the illusion is there.

A few more asks for me:
1. Could we have a consistent footer? Currently the Docs pages don't have any footer at all.
2. I think, given it's a whole micro-site, this deserves it's own top-level menu item in the header rather than being buried under the About menu.
3. I still don't understand the need for `src/content/docs/docs`. Can't it all be under `src/content/docs/` and avoid that double `doc/docs`?

max-ostapenko:
@tunetheweb 'docs/docs/' is Starlights artifact.
It expects a collection under 'src/content/docs/', but as our website is not just docs, keeping the documentation right under 'docs/' will conflict with the rest of Astro pages.

And Starlight doesn't have a feature to define prefix.

max-ostapenko:
@tunetheweb with unified footer and header now

